PR #24, opened 2019-06-03
Reservoir simulation time series

WestfalNamur:
Changes made in _summary_stats.py
- Added 
    multiple ensembles,
    history-vectors (if existing with uncertainty),
    multiple plots for statatistics (one per ensemble)
- Smaller bug fixes
- No other files modified.


WestfalNamur:
I fixed the 1st err. in Travis. The 2nd "Plugin 'pytest_dash' could not be loaded:" seems not be related to _summary_stats.py. 
I will wait for a review and the possibility before commiting again.

webviz_subsurface/containers/_summary_stats.py
```diff
@@ -7,41 +7,69 @@
 from webviz_config.webviz_store import webvizstore
 from webviz_config.common_cache import cache
 from ..datainput import scratch_ensemble
+from plotly.colors import DEFAULT_PLOTLY_COLORS
+import itertools
 
 
 class SummaryStats:
```
HansKallekleiv: All containers should now inherit the WebvizContainer class.
class SummaryStats(WebvizContainer)

webviz_subsurface/containers/_summary_stats.py
```diff
@@ -7,41 +7,69 @@
...
-            ensemble,
+            ensembles,
+            column_keys=None,
             sampling: str = 'monthly',
-            title: str = 'Simulation time series'):
+            title: str = 'Simulation time series',
+            history_uncertainty: bool = False):
 
         self.title = title
         self.dropwdown_vector_id = 'dropdown-vector-{}'.format(uuid4())
```
HansKallekleiv: - Rename to `self.dropdown_vector_id` ?
- Enough to create one uid `self.uid = f'{uuid4()}'`
- Change all string formatting to  f-strings e.g. `f'dropdown-vector-{self.uid}'`

webviz_subsurface/containers/_summary_stats.py
```diff
@@ -7,41 +7,69 @@
...
+            list(
+                get_summary_data(
+                    ensemble_paths=self.ensemble_paths,
+                    sampling=self.sampling,
+                    column_keys=self.column_keys) .drop(
+                    columns=[
+                        'DATE',
+                        'REAL',
+                        'ENSEMBLE']) .columns))
+        self.smry_vector_columns = tuple([col for col in self.smry_columns
```
HansKallekleiv: `col_endswith("H"'`) is not correct as this doesn't differentiate vectors e.g. per well.
E.g. the vector `GGOR:WELLH` should not be filtered.
=> One solution could be to split on `:` and filter on the first element only

webviz_subsurface/containers/_summary_stats.py
```diff
@@ -52,69 +80,199 @@ def layout(self):
...
-                                     'sampling': self.sampling,
-                                     'statistics': False}]),
-                (get_summary_data, [{'ensemble_path': self.ensemble_path,
-                                     'sampling': self.sampling,
-                                     'statistics': True}])]
+        return [(get_summary_data,
+                 [{'ensemble_paths': self.ensemble_paths,
+                   'column_keys': self.column_keys,
+                   'sampling': self.sampling,
+                   'smry_history_columns': self.smry_history_columns,
```
HansKallekleiv: This function does not run currently as `get_summary_data` does not take all these arguments

webviz_subsurface/containers/_summary_stats.py
WestfalNamur: Pseudocode now:
    lstH = [i + 'H' for i in vctrs if i + 'H' in vctrs]
    lst = [i for i in vctrs if i not in lstH]

PR #36, opened 2019-06-18
Remove user provided container titles

### LINKED ISSUE #34: Remove html Title argument from containers
Some containers has a `title` argument as input which renders a dcc.H1 element.
This should instead be handled by markdown outside the container to allow more flexibility in container placement and styling.

HansKallekleiv:
closes #34

### DIFF
--- a/webviz_subsurface/containers/_history_match.py
+++ b/webviz_subsurface/containers/_history_match.py
@@ -19,11 +19,9 @@ This container visualizes the quality of the history match.
 * `title`: Optional title for the container.
 '''
 
-    def __init__(self, container_settings, ensembles, observation_file: Path,
-                 title: str = 'History Match'):
+    def __init__(self, container_settings, ensembles, observation_file: Path):
 
         self.observation_file = observation_file
-        self.title = title
         self.ens_paths = tuple((ens,
                                 container_settings['scratch_ensembles'][ens])
                                for ens in ensembles)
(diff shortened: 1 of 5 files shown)


webviz_subsurface/containers/_history_match.py
```diff
@@ -99,7 +97,6 @@ def _iterations_to_dict(iterations, labels):
     @property
     def layout(self):
         return html.Div([
-                  html.H2(self.title),
                   wsc.HistoryMatch(id=self.hm_id, data=self.hm_data)
                ])
```
anders-kiaer: Can this now be changed to
`return wsc.HistoryMatch(id=self.hm_id, data=self.hm_data)`
?

webviz_subsurface/containers/_subsurface_map.py
```diff
@@ -40,7 +39,6 @@ def __init__(self, container_settings, ensemble, map_value: str,
     @property
     def layout(self):
         return html.Div([
-                   html.H2(self.title),
                    Map(id=self.map_id, data=self.map_data)
                ])
```
anders-kiaer: Can this now be changed to
`return Map(id=self.map_id, data=self.map_data)`
?

webviz_subsurface/containers/_history_match.py
HansKallekleiv: They could, but I think having them wrapped in a div is something we will need later on. To control styling and sizing and such. I think we can just keep the div's there for now.

PR #38, opened 2019-06-21
Adding fieldgains to time-series plot

WestfalNamur:
- adding fieldgain traces: compare to ensembles with each otther
- datainput functions operate on ensemleset
- change layout to plot selectors next to plot


WestfalNamur:
1. ModuleNotFoundError: No module named 'fmu'  during Travis-testing. 
 Why can't fmu be imported here?
2. May need another testing method like unittest to avoid assert and this the code-quality check issues?

WestfalNamur:
Currently missing the portable option to include fieldgains/delta-values.
Adding an enesmble_set_name as an inputin add_webvizstore allows use of portable option but it is missing the fieldgains option.
This is because missing of get_fieldgain(). Since this is a combination of two ensembles we would need to loop over every possbile combination and add it to the store or storing the whole ensemble-object and operate on it (would probalby be to big and missing one of he advantages of webviz #filteringbycolumn_keys). Or we could make a new function to calculate and compare two summary_data sets. This would mean that we would have some post-processing logic in webviz. Maybe this is a scope of fmu-ensemble?

WestfalNamur:
Closed and replaced by _time_series container

PR #45, opened 2019-08-06
Use only numerical columns for correlation matrix

### LINKED ISSUE #44: Parameter Distribution correlation matrix does not render when first data point is undefined
Plotly Heatmap returns the following console error when first data point provided has Nan/Null value.
```javascript
react-dom@16.8.6.min….0&m=1565085890:117 TypeError: Cannot read property '0' of undefined
    at u (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at d (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at e.exports (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at Object.e.exports [as calc] (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at b (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at Object.v.doCalcdata (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at r.plot (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at Object.l.syncOrAsync (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at Object.r.react (plotly-1.48.3.min.js….0.0&m=1565085961:7)
    at t.value (Graph.react.js:108)
```

Easiest solution is to just filter non numerical column from the parameters dataframe before plotting.
Alternatively nan values must be given a value and added as a separate heatmap trace.

HansKallekleiv:
closes #44


webviz_subsurface/containers/_parameter_distribution.py
```diff
@@ -319,7 +320,7 @@ def render_scatter(ens1, x_col, ens2, y_col, color, density):
 @cache.memoize(timeout=cache.TIMEOUT)
 def render_matrix(ensemble_path):
 
-    data = get_parameters(ensemble_path)
+    data = get_parameters(ensemble_path).select_dtypes([np.number])
```
anders-kiaer: Are we sure `get_parameters` will always return `np.number` as numerical data type for numerical values?

Is it possible + more robust to e.g. first use [`to_numeric`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html) with argument `errors='coerce'` followed by [`dropna`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.dropna.html)? That would also make `import numpy as np` unnecessary.

HansKallekleiv: Yes, seems possible and more robust :)
Fixed

PR #58, opened 2019-08-28
Fix parameter distribution heatmap

WestfalNamur:
- Dropnan after calculating correlation matrix to drop nan values resulting
  from undefined (comparing constants).


WestfalNamur:
Should we wait until https://github.com/equinor/webviz-subsurface/pull/60 is merged? 
Then I can rebase and maybe sqash the changes in this branch into on commit?

PR #61, opened 2019-08-29
Use Plotly graph from webviz-core-components

HansKallekleiv:
Following https://github.com/equinor/webviz-core-components/pull/12 replace use of `dash_core_components.Graph()` with `webviz_core_components.Graph()`


webviz_subsurface/containers/_inplace_volumes.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 import dash_html_components as html
```
anders-kiaer: There are some docstrings in this file refering to `dcc.Graph`. Replace with `wcc.Graph`?

PR #75, opened 2019-09-11
Optionally visualize non-varying parameters in ParameterDistribution container

### LINKED ISSUE #62: Parameter distribution dropdown
In practice, some parameter names are so long that you only see the same names (i.e. the prefix).

image

### LINKED ISSUE #68: ParameterDistribution: Inconsistency in parameters used in correlation plot and scatterplot/dropdowns
Currently parameters with constant values are filtered out before the correlation plot is rendered, but they are still available for selection in the dropdown menus.
Currently a ValueError is raised if the dropdowns are set to parameters that are not available in the correlation plot.

We need to either remove constant parameters all together or add an exception block for the ValueError.

HansKallekleiv:
Adds config file optional boolean argument `drop_constants` to `ReservoirSimulationTimeSeries`
Default is `True` which will drop all parameters that are not varying.
Fixes  #68 
Also some small styling changes that fixes #62

### DIFF
--- a/webviz_subsurface/containers/_parameter_distribution.py
+++ b/webviz_subsurface/containers/_parameter_distribution.py
@@ -28,12 +28,15 @@ This container shows parameter distribution as histogram,
 and correlation between the parameters as a correlation matrix.
 
 * `ensembles`: Which ensembles in `container_settings` to visualize.
+* `drop_constants`: Drop constant parameters (Default: true)
 '''
 
-    def __init__(self, app, container_settings, ensembles):
+    def __init__(self, app, container_settings,
+                 ensembles, drop_constants: bool = True):
 
         self.ensembles = {ens: container_settings['scratch_ensembles'][ens]
                           for ens in ensembles}
+        self.drop_constants = drop_constants
         self.uid = f'{uuid4()}'
         self.p1_drop_id = f'p1-dropd-{self.uid}'
         self.p2_drop_id = f'p2-dropd-{self.uid}'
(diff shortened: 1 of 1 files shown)


webviz_subsurface/containers/_parameter_distribution.py
```diff
@@ -28,12 +29,15 @@ class ParameterDistribution(WebvizContainer):
 and correlation between the parameters as a correlation matrix.
 
 * `ensembles`: Which ensembles in `container_settings` to visualize.
+* `drop_constants`: Drop constant parameters (Default: true)
```
anders-kiaer: Could drop the `(Default: true)` as that is added to the generated documentation automatically from the function signature.

webviz_subsurface/containers/_parameter_distribution.py
```diff
@@ -309,18 +322,26 @@ def render_scatter(ens1, x_col, ens2, y_col, color, density):
...
+    they are undefined.
+    Passing tuple or list to drop on multiple axes is deprecated since
+    version 0.23.0. Therefor split in 2x .dropnan()
+    '''
     data = get_parameters(ensemble_path).apply(pd.to_numeric, errors='coerce')\
                                         .dropna(how='all', axis='columns')
+    if drop_constants:
+        return data.corr().dropna(axis='index', how='all') \
+                               .dropna(axis='columns', how='all')
+    return data.corr()
```
anders-kiaer: Possible to get away with one `return`? E.g.
```python
return data.corr() if not drop_constants \
    else data.corr().dropna(axis='index', how='all') \
    .dropna(axis='columns', how='all')
```

Issue #82, opened 2019-09-16
Input configuration for surfaces


HansKallekleiv:
Relates to #69 

In order to facilitate easy selection and visualization of surface data from an ensemble some functionality is needed to show which data is available.

Work is currently ongoing to resolve this (@perolavsvendsen) , but until then we need a band-aid solution that will allow users to visualise these data. Ideally we should be able to make something that later can be modified.

Below is a suggestion using a user-provided configuration file and a ui widget to select data. 

The suggestion has many limitations and  **is just meant to get the discussion started**

@jcrivenaes
@anders-kiaer 
@berland

Any feedback is welcome.

### How data is stored
Currently surface is stored as files in an ensemble on the following relative path
`share/results/maps`
Individual surface grids are stored within that path as:
`surface_name--surface_property--surface_timestep.gri`

The name components can be described as:
**Surface grid** a 3-dimensional collection of points, x, y, z, where x,y are coordinates while z is an arbritrary property.

**Surface name** The name of the entity that the surface represents. This could be e.g. a seismic interpreted reflector,  a geological interpreted isochore or a fluid contact.

**Surface property** The z-value attribute of the surface grid. This could be e.g. depth values, time values, mean porosity or saturation.

**Surface timestep** A surface property can have a time dimension, e.g. pressure changes over time

- *Surface names* and *Surface timesteps* must be ordered.
- Different *surface properties* can have different lists of names and timesteps
- Different *surface properties* does not have to be ordered

A simple way to display this information can be:
```yaml

#Configuration file for surfaces
#surface_property:
#  surface_names:
#    - some name
#    - another name
#  surface_timestep:
#    - some dat
#    - another date

average_permx:
  names: 
    - all
    - upperreek
    - midreek
    - lowerreek

average_pressure:
  names:
    - all
    - upperreek
    - midreek
    - lowerreek
  dates:
    - 20000101
    - 20010601
    - 20030101
    - 20010601_20000101
    - 20030101_20000101
    - 20030101_20010601

ds_extracted_horizons:
  names:
    - topupperreek
    - topmidreek
    - toplowerreek
    - baselowerreek

```

This can be then be made into a UI widget that returns the selected surface object.
![surface_selector0](https://user-images.githubusercontent.com/16436291/64949907-83c75800-d87a-11e9-9d51-7be0e50e18b0.gif)

The returned object can then e.g. be used to retrieve, calculate and visualize data through an API or from file based storage:

[...]

PR #83, opened 2019-09-16
Input configuration for surfaces

HansKallekleiv:
Relates to #82 

Basic usage:
Config file:
```yaml
title: Surface selector test

container_settings:
  scratch_ensembles:
    history_match_prior: /scratch/fmu/hakal/9_r001_reek/realization-*/iter-0
    history_match_update: /scratch/fmu/hakal/9_r001_reek/realization-*/iter-4
    design_matrix: /scratch/fmu/hakal/2_r001_reek_fullmatrix/realization-*/iter-0
pages:
 - title: surface test
   content:
    - container: surf.SurfaceSelectorTest
      ensembles:
        - history_match_prior
        - history_match_update
        - design_matrix
```

Example container:
```python
import json
from dash.dependencies import Input, Output
from webviz_config.containers import WebvizContainer
from webviz_subsurface.private_containers._surface_selector import (
    SurfaceSelector,
)
from webviz_subsurface.datainput import get_realizations, find_surfaces


class SurfaceSelectorTest(WebvizContainer):
    def __init__(self, app, container_settings, ensembles):
        self.ensemble_paths = tuple(
            (ensemble, container_settings["scratch_ensembles"][ensemble])
            for ensemble in ensembles
        )
        realizations = get_realizations(
            self.ensemble_paths, ensemble_set_name="EnsembleSet"
        )
        surface_context = find_surfaces(self.ensemble_paths)
        self.selector = SurfaceSelector(app, surface_context, realizations)
        self.set_callbacks(app)

    @property
    def layout(self):
        return html.Div(children=[self.selector.layout, html.Pre(id="pre")])

    def set_callbacks(self, app):
        @app.callback(
            Output("pre", "children"),
            [Input(self.selector.storage_id, "children")],
        )
        def _test(data):
            return json.dumps(json.loads(data), indent=2)


		
```


webviz_subsurface/datainput/_fmu_input.py
```diff
@@ -23,3 +25,66 @@ def load_parameters(
...
+        dates:
+            - some_date
+            - another_date
+    """
+
+    files = []
+    for _, path in ensemble_paths:
+        files.extend(glob.glob(os.path.join(path, "share", "results", "maps", suffix)))
+
+    files = [os.path.basename(f.split(".gri")[0]) for f in files]
```
anders-kiaer: This takes out the file name without file extension? See [pathlib.PurePath.stem](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem), if `f` is a `Path` you can maybe replace
```python
os.path.basename(f.split(".gri")[0])
```
with `f.stem`.

webviz_subsurface/datainput/_fmu_input.py
```diff
@@ -23,3 +25,66 @@ def load_parameters(
...
+    files = [f.split("--") for f in files]
```
anders-kiaer: This could perhaps be joined with the line before, i.e. `f.stem.split("--")`.

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,564 @@
...
+
+    @staticmethod
+    def read_config(config):
+        """Reads config file either from a yaml provided file or from a dict"""
+        if isinstance(config, str):
+            return yaml.safe_load(open(config, "r"))
+        elif isinstance(config, dict):
+            return config
+        else:
+            raise IOError("Config must be a dictionary of a yaml file")
```
anders-kiaer: `TypeError` or `ValueError` instead of `IOError`?

webviz_subsurface/datainput/_fmu_input.py
```diff
@@ -23,3 +25,66 @@ def load_parameters(
...
+            data["name"] = f[0]
+        except IndexError:
+            data["name"] = None
+        try:
+            data["date"] = f[2]
+        except IndexError:
+            data["date"] = None
+        arr.append(data)
+
+    df = pd.DataFrame(arr)
```
anders-kiaer: Could line 64-81 be simplified/remvoed by utilising that a dataframe can be generated directly from list of lists (even of varying length):
```python
>>> pandas.DataFrame([[1, 2, 3], [4, 5]], columns=['A', 'B', 'C'])
   A  B    C
0  1  2  3.0
1  4  5  NaN
>>> 
```

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,564 @@
...
+from dash.exceptions import PreventUpdate
+import dash_html_components as html
+import dash_core_components as dcc
+
+
+class SurfaceSelector:
+    """### Surface Selector
+
+Creates a widget to select surfaces from a configuration file.
+The current selection are stored in a dcc.Store object that can
```
anders-kiaer: are --> is?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,564 @@
...
+def format_date(date_string):
+    d = str(date_string)
+    if len(d) == 8:
+        d = datetime.strptime(d, "%Y%m%d").strftime("%m/%d/%Y")
+    if len(d) == 17:
+        begin = d[0:8]
+        end = d[9:17]
+        d = f"({datetime.strptime(begin, '%Y%m%d').strftime('%m/%d/%Y')})-\
+              ({datetime.strptime(end, '%Y%m%d').strftime('%m/%d/%Y')})"
+    return d
```
anders-kiaer: - On purpose to change it from `YYYYMMDD` to `MM/DD/YYYY`?
- If we are doing some formatting here, we could perhaps simplify it such that the container optionally only shows e.g. `February 2020` instead of full date (typically knowing the month is maybe enough wrt. the time resolution of the physical processes in the subsurface). Either as a boolean argument to the container (show day of month or not), or some logic in the container (if delta between dates of same month and year, or if two neighbouring dates are in same month + year --> show day)

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,589 @@
...
+
+def format_date(date_string):
+    """Reformat date string for presentation
+    20010101 => Jan 2001
+    20010101_20010601 => (Jan 2001) - (June 2001)
+    20010101_20010106 => (01 Jan 2001) - (06 Jan 2001)"""
+    d = str(date_string)
+    if len(d) == 8:
+        d = datetime.strptime(d, "%Y%m%d").strftime("%b %Y")
+    if len(d) == 17:
```
anders-kiaer: `elif`, and then perhaps raise a `ValueError` in a final `else` or fallback to return `date_string`.

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,589 @@
...
+    else:
+        return current_value
+
+
+def format_date(date_string):
+    """Reformat date string for presentation
+    20010101 => Jan 2001
+    20010101_20010601 => (Jan 2001) - (June 2001)
+    20010101_20010106 => (01 Jan 2001) - (06 Jan 2001)"""
+    d = str(date_string)
```
anders-kiaer: Do we need to give `date_string` a new name?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,589 @@
...
+    if len(d) == 8:
+        d = datetime.strptime(d, "%Y%m%d").strftime("%b %Y")
+    if len(d) == 17:
+        begin = d[0:8]
+        end = d[9:17]
```
anders-kiaer: Instead `[begin, end] = [datetime.strptime(date, '%Y%m%d') for date in d.split('-')]`?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,589 @@
...
+        if begin[0:4] == end[0:4] and begin[4:6] == end[4:6]:
```
anders-kiaer: ...because then this can be something like 
```python
if begin.year == end.year and begin.month == end.month:
```

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,589 @@
...
+
+    def names_in_attr(self, attr):
+        return self._configuration[attr].get("names", None)
+
+    def dates_in_attr(self, attr):
+        dates = self._configuration[attr].get("dates", None)
+        if dates:
+            return [d for d in dates]
+        else:
+            return None
```
anders-kiaer: `return dates if dates else None`?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+        self.attr_wrapper_id = f"{uuid}-attr-wrapper"
+        self.name_wrapper_id = f"{uuid}-name-wrapper"
+        self.date_wrapper_id = f"{uuid}-date-wrapper"
+        self.ens_wrapper_id = f"{uuid}-ens-wrapper"
+        self.real_wrapper_id = f"{uuid}-real-wrapper"
+        self.aggreal_id = f"{uuid}-aggreal"
+        self.sens_name_id = f"{uuid}-sens-name-id"
+        self.sens_case_id = f"{uuid}-sens-case-id"
+        self.sens_name_wrapper_id = f"{uuid}-sens-name-wrapper-id"
+        self.sens_case_wrapper_id = f"{uuid}-sens-case-wrapper-id"
```
anders-kiaer: Some of the IDs are not used in callbacks, then they can perhaps be removed altogether? Both here + in the component they are currently used in?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+    def aggregations(self):
+        return ["mean", "stddev", "min", "max", "p10", "p90"]
+
+    @property
+    def show_dropdown_style(self):
+        return {"display": "grid"}
+
+    @property
+    def hide_dropdown_style(self):
+        return {"visibility": "hidden"}
```
anders-kiaer: The `@property` decorated functions with constant return value, should they instead simply be either class or instance constants instead?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+                    "senscase": sens_case,
+                    "all_senscases": all_senscases,
+                }
+            )
+
+
+def prev_value(current_value, options):
+    try:
+        index = options.index(current_value)
+    except ValueError:
```
anders-kiaer: In which scenario does this happen (i.e. that the current value is not among the options)?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+        index = None
+    if index > 0:
+        return options[index - 1]
+    else:
+        return current_value
```
anders-kiaer: Instead of
```
if index > 0:
    return options[index - 1]
else:
    return current_value
```
consider
`return options[max(0, index - 1)]`

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+
+def next_value(current_value, options):
+    try:
+        index = options.index(current_value)
+    except ValueError:
+        index = None
+    if index < len(options) - 1:
+        return options[index + 1]
+    else:
+        return current_value
```
anders-kiaer: Same questions as for  `prev_value`.

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+
+
+def format_date(date_string):
+    """Reformat date string for presentation
+    20010101 => Jan 2001
+    20010101_20010601 => (Jan 2001) - (June 2001)
+    20010101_20010106 => (01 Jan 2001) - (06 Jan 2001)"""
+    date_string = str(date_string)
```
anders-kiaer: Input argument `date_string` is not string already?

webviz_subsurface/private_containers/_surface_selector.py
```diff
@@ -0,0 +1,586 @@
...
+    if len(date_string) == 8:
+        return datetime.strptime(date_string, "%Y%m%d").strftime("%b %Y")
+    elif len(date_string) == 17:
+        [begin, end] = [
+            datetime.strptime(date, "%Y%m%d") for date in date_string.split("_")
+        ]
+        if begin.year == end.year and begin.month == end.month:
+            return f"({begin.strftime('%d %b %Y')})-\
+              ({end.strftime('%d %b %Y')})"
```
anders-kiaer: If you use `%-d` instead of `%d`, there will not be a leading zero for the days 1-9 in each month.

webviz_subsurface/private_containers/_surface_selector.py
HansKallekleiv: I could only find one that is not used. Removed that one :)

webviz_subsurface/private_containers/_surface_selector.py
HansKallekleiv: The options in one dropdown are updated based on selections in another dropdown.
The callback are handling both setting new options and setting the current value from existing options or new options. 
When this function is called it could be that the options have changed.

PR #86, opened 2019-09-19
New ParameterDistribution container

HansKallekleiv:
- Renamed existing ParameterDistribution to ParameterCorrelation
- Added new ParameterDistribution container

The new container shows a single parameter in a Plotly Express  histogram with a marginal boxplot on top.


anders-kiaer:
Looks good! 🍾 💃 

Since apparently `git mv` + create a new file with the old name confuses the github diff, base the changes here after merge of #87 (in order to make the diff smaller)?

PR #88, opened 2019-09-20
Added parameter distribution container

HansKallekleiv:
Added ParameterDistribution container that shows multiple ensembles of model parameters in histogram/boxplot

### DIFF
--- a/webviz_subsurface/containers/__init__.py
+++ b/webviz_subsurface/containers/__init__.py
@@ -35,11 +35,11 @@ from ._reservoir_simulation_timeseries import ReservoirSimulationTimeSeries
 
 
 __all__ = ['ParameterDistribution',
+           'ParameterCorrelation',
            'DiskUsage',
            'SubsurfaceMap',
            'HistoryMatch',
            'Intersect',
            'MorrisPlot',
            'InplaceVolumes',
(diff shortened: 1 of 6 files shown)


webviz_subsurface/containers/__init__.py
```diff
@@ -33,11 +34,12 @@
...
+__all__ = ['ParameterDistribution',
            'DiskUsage',
            'SubsurfaceMap',
            'HistoryMatch',
            'Intersect',
            'MorrisPlot',
            'InplaceVolumes',
-           'ReservoirSimulationTimeSeries']
+           'ReservoirSimulationTimeSeries',
+           'ParameterCorrelation']
```
anders-kiaer: Unnecessary diff in the end?

HansKallekleiv:
> * Should/could `_parameter_correlation` also use the `_fmu_input.load_parameters` introduced here?

Yes. Perhaps better to handle in a separate PR? Could also attempt to make some of the other containers use this code. Currently I believe the same ensemble / ensemblesets are loaded multiple times in different containers.

HansKallekleiv:
> * Add option to normalize to percent instead of count?

Skipping this one unless seen as important. The plot resets when changing mode. Would require adding uirevision to the figure layout...

PR #105, opened 2019-10-03
Guided tours for containers

HansKallekleiv:
Added tours for:
- InplaceVolumesOneByOne
- ReservoirSimulationTimeSeriesOneByOne

### DIFF
--- a/webviz_subsurface/containers/_inplace_volumes_onebyone.py
+++ b/webviz_subsurface/containers/_inplace_volumes_onebyone.py
@@ -114,6 +114,7 @@ Requires ensembles with `SENSNAME`and `SENSCASE` information.
         self.tornadowrapper_id = f"tornadowrapper-{uuid}"
         self.source_id = f"source-{uuid}"
         self.ensemble_id = f"ensemble-{uuid}"
+        self.filter_selectors_id = f"filter-selectors-{uuid}"
         self.selectors_id = {x: f"{x}{uuid}" for x in self.selectors}
 
     def add_webvizstore(self):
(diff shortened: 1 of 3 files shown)


webviz_subsurface/containers/_inplace_volumes_onebyone.py
```diff
@@ -139,6 +140,55 @@ def add_webvizstore(self):
...
+        return [
+            {
+                "id": self.graph_id,
+                "content": "The chart shows inplace volumetrics results. ",
+            },
+            {
+                "id": self.table_id,
+                "content": "The table shows statistics per sensitivity. \
+                            Rows can be filtered by searching, and sorted by \
+                            clicking on a column header. ",
```
anders-kiaer: - Using `\` gives strings with a lot of spaces (since the indentation of next line is added as spaces). Visually not a problem I guess here since extra spaces are not rendered in html, but perhaps we should change these multilines to e.g.
    ```python
    "content" = ( "The table shows statistics per sensitivity. "
                        "Rows can be filtered by searching, and sorted by "
                        "clicking on a column header."),
   ```
- Unnecessary space at end of last line.
- Consider replacing _sensitivity_ with _sensitivity parameter_ or _parameter_? (Or keep as is - I'm not sure)

webviz_subsurface/containers/_inplace_volumes_onebyone.py
```diff
@@ -139,6 +140,55 @@ def add_webvizstore(self):
...
+            {
+                "id": self.plot_type_id,
+                "content": "Controls the type of the visualized chart. \
+                            Per realization shows bars per realization, \
+                            while the boxplot shows the range per sensitivity. ",
+            },
+            {
+                "id": self.tornadowrapper_id,
+                "content": "Tornado plot for the currently displayed data. \
+                            Differences references can be set and sensitivities \
```
anders-kiaer: Strange start of sentence?

webviz_subsurface/containers/_inplace_volumes_onebyone.py
```diff
@@ -139,6 +140,55 @@ def add_webvizstore(self):
...
+            },
+            {
+                "id": self.ensemble_id,
+                "content": "If several ensembles are available, the active ensemble \
+                            can be selected here. ",
+            },
+            {
+                "id": self.source_id,
+                "content": "If volumes have been calculated for different grids. \
+                            E.g. geogrid and eclipsegrid, the active grid can be selected here. ",
```
anders-kiaer: Maybe we could make an issue out of this: The source selector does not appear if only calculated for one grid?

webviz_subsurface/containers/_inplace_volumes_onebyone.py
```diff
@@ -139,6 +140,55 @@ def add_webvizstore(self):
...
+            },
+            {
+                "id": self.filter_selectors_id,
+                "content": "Filter on different combinations of e.g. zones, facies and regions \
+                            (The options will vary dependent on what was included \
+                            in the calculation. ",
```
anders-kiaer: Since the `tour_steps` property has access to `self`, the tooltips could actually include the selectors included in each case. :smiley: Maybe not very important, if difficult to achieve.

webviz_subsurface/containers/_reservoir_simulation_timeseries_onebyone.py
```diff
@@ -102,12 +102,33 @@ def __init__(
...
         self.ensemble_id = f"ensemble-{uuid}"
         self.table_id = f"table-{uuid}"
         self.graph_id = f"graph-{uuid}"
         self.graph_wrapper_id = f"graph-wrapper-{uuid}"
+        self.tornadowrapper_id = f"tornadowrapper-{uuid}"
+
+    @property
+    def tour_steps(self):
+        return [
+            {"id": self.smry_col_id, "content": "Displayed summary vector"},
```
anders-kiaer: Use a different word than `summary vector`? `time series`?

webviz_subsurface/containers/_inplace_volumes_onebyone.py
HansKallekleiv: I think it should be there even for one grid. Nice to know which grid was used for the calculation.

webviz_subsurface/containers/_inplace_volumes_onebyone.py
HansKallekleiv: It shouldn't be too difficult to achieve, but I'm not sure how much more information it would give?
I think zone, region, facies and licenses (the available filters) are fairly well known.
For geologists atleast :)

PR #118, opened 2019-11-06
Adding support for 'find_files' for UNSMRY files

kjetilbjorke:
Added support for discovery of `UNSMRY` files in non standard location (or multiple `UNSMRY` files in same folder).
Usage in `yaml` config file:
```
container_settings:
  scratch_ensembles:
    iter-0: /path/to/scratch/case/realization-*/iter-0
    iter-0_ix: /path/to/scratch/case/realization-*/iter-0
  unsmry_files: 
    iter-0: eclipse/model/*.UNSMRY
    iter-0_ix: ix/model/*.UNSMRY

### DIFF
--- a/webviz_subsurface/containers/_reservoir_simulation_timeseries.py
+++ b/webviz_subsurface/containers/_reservoir_simulation_timeseries.py
@@ -63,7 +63,8 @@ class ReservoirSimulationTimeSeries(WebvizContainerABC):
         )
         self.unsmry_files = tuple(
             (ensemble, container_settings["unsmry_files"][ensemble])
-            for ensemble in ensembles if "unsmry_files" in container_settings
+            for ensemble in ensembles
+            if "unsmry_files" in container_settings
         )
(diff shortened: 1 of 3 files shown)


anders-kiaer:
Thanks @kjetilbjorke! This looks like good addition to the feature set.

Wondering if we should merge @HansKallekleiv test data set first, get those tests to run, and then add this one afterwards. Perhaps also after @HansKallekleiv complete rewrite of simulation time series container.

HansKallekleiv:
Thanks @kjetilbjorke . This container is completely rewritten now as we needed to improve the performance and add some more features (display more than one vector, show observations, etc).

Although discovery of `UNSMRY` files in non standard locations has not been added, it is now possible to initialize the container from csv files.
This will allow for any non-standard input as long as the file format of the csv file follows the standard. 

Closing this PR without merge, but feel free to add/request this functionality in the new container if you still see the need!

PR #125, opened 2019-11-11
SEG-Y viewer

HansKallekleiv:
Resolves #123.

PR #126, opened 2019-11-11
Segyviewer(Leaflet)

HansKallekleiv:


anders-kiaer:
We will close this as of now, and prefer #125 instead. Might be worth revisiting when these are solved:
- https://github.com/equinor/webviz-subsurface-components/issues/83
- https://github.com/equinor/webviz-subsurface-components/issues/82
- https://github.com/equinor/webviz-subsurface-components/issues/37
- https://github.com/equinor/webviz-subsurface-components/issues/33

PR #158, opened 2019-11-27
Identify private modules

### LINKED ISSUE #145: Rename datainput submodule to _datainput
...in order to communicate that it is used internally by `webviz-subsurface` containers.

They might become  public API at some point when more matured - but for now they will change much more rapidly than the public container APIs/arguments.

anders-kiaer:
Closes #145.


anders-kiaer:
It was intentional, but when you say it, maybe we when doing this change should remove all imports in `[...]/_datainput/__init__.py` and let imports be directly from the submodules?

HansKallekleiv:
I'd say just remove it all from `__init__`

PR #171, opened 2019-12-03
Dashboard for parameter and response correlation

HansKallekleiv:
Added a dashboard to do basic correlation between input parameters and any response.
The responses are given as a user-provided csv file.

The dashboard consists of :
- pearson/spearman correlation between a selected response and all parameters (horizontal bar chart)
- Distribution for individual parameters / responses (scatterplot / histogram) updated from click-event on correlation plot
- Optional response filters. E.g. filter on `DATE` if the response is a time series.


webviz_subsurface/containers/_parameter_response_correlation.py
```diff
@@ -0,0 +1,622 @@
...
+class ParameterResponseCorrelation(WebvizContainerABC):
+    """### Parameter and response correlation
+
+This container shows correlation between numerical input parameters and responses.
+
+Input can be given either as:
+
+- Aggregated csv files for parameters and responses,
+- An ensemble name defined in shared_settings and a local csv file for responses
+stored per realizations.
```
anders-kiaer: - Remove empty line 25, and instead indent the two lines starting with `-` for readability?
- `shared_settings` instead of shared_settings.

PR #207, opened 2020-01-24
Small fixes to surface calculations

HansKallekleiv:
- Calculate bounds from unrotated surface
- Fill masked np array with nan values before dumping to file


webviz_subsurface/plugins/_well_cross_section_fmu.py
```diff
@@ -591,7 +592,7 @@ def surface_to_json(surface):
             "rotation": surface.rotation,
             "xinc": surface.xinc,
             "yinc": surface.yinc,
-            "values": surface.values.tolist(),
+            "values": values.tolist(),
```
anders-kiaer: Can all the three changed lines be comineed into only changing the last one?
```python
"values": surface.values.filled(np.nan)
```

PR #232, opened 2020-02-19
Test Travis caching of testdata

anders-kiaer:


anders-kiaer:
Did not look to give increased CI performance wrt. testdata caching.

Issue #179, opened 2019-12-06
Replace scientific notation with SI prefixes


HansKallekleiv:
M (mega), G (giga), etc...

PR #275, opened 2020-03-05
WIP: Tornado table

### LINKED ISSUE #236: Tornadoplot: Add tornado table
These are numbers that are often reported and would be useful to have in a table.
Perhaps same info as you already have in the plot, but open for other suggestions.

Example of what it could look like:

**Sensitivity compared to reference `<reference average name>` with average value `<ref avg value>`** 

Sensitivity | lowcase | highcase | delta low  | delta high | true low | true high | low #reals | high #reals
------------|----------|-----------|------------|-------------|-----------|----------|------------|----------
rms_seed | p90 | p10 | -3.42 | 3.24 | 1.114e+08 | 1.192e+08 | 10 | 10 
gwc | shallow | deep | -12.12 | 13.89 | 1.014e+08 | 1.315e+08 | 10 | 9

* Sorted in the same way as tornado plot, highest sensitivity on top.
* delta low always lower than delta high, and casenames flipped also if values are flipped (like for the labels on the plot)
* delta low/delta high are delta values with respect to reference. Could be either percentage or absolute (i.e. the same as for calculated in plot). 
* true low and true high are values where reference is not subtracted
* reals low and reals high are how many realisations that went into the low and high calculations

HansKallekleiv:
Closes #236

### DIFF
--- a/webviz_subsurface/_private_plugins/tornado_plot.py
+++ b/webviz_subsurface/_private_plugins/tornado_plot.py
@@ -5,6 +5,7 @@ import pandas as pd
 import dash
 from dash.dependencies import Input, Output
 from dash.exceptions import PreventUpdate
+import dash_table
 import dash_html_components as html
 import dash_core_components as dcc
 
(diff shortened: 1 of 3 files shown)


HansKallekleiv:
Dependent on https://github.com/equinor/webviz-subsurface/issues/179

anders-kiaer:
Consider if data download is enough?

PR #280, opened 2020-03-10
Number formatting / SI prefix

### LINKED ISSUE #179: Replace scientific notation with SI prefixes

asnyv:
Closes #179


webviz_subsurface/_abbreviations/__init__.py
```diff
@@ -29,3 +39,61 @@ def simulation_vector_description(vector):
         )
 
     return description
+
+
+def si_prefixed(
```
asnyv: Should we consider to move logic out of this __init__ file to separate modules (both simulation_vector_description and si_prefixed)?

anders-kiaer: Yes, that might be a good idea, now that `_abbreviations` is expanding. 👍

webviz_subsurface/_abbreviations/volume_units.json
```diff
@@ -0,0 +1,26 @@
...
+  "STOIIP_OIL": "Sm³",
+  "STOIIP_GAS": "Sm³",
+  "STOIIP_TOTAL": "Sm³",
+  "GIIP_OIL": "Sm³",
+  "GIIP_GAS": "Sm³",
+  "GIIP_TOTAL": "Sm³",
+  "RECOVERABLE_OIL": "Sm³",
+  "RECOVERABLE_GAS": "Sm³",
+  "RECOVERABLE_TOTAL": "Sm³"
+}
```
asnyv: Is it too risky to assume units? Or should e.g. user be able to overwrite this file by linking to another file? (if so, plugin by plugin, or for the whole app?). These units are also just my "best guess", would appreciate if e.g. @HansKallekleiv could take a look and see if they are correct. All units are volumes as far as I can understand, so is a question of standard conditions or not.

anders-kiaer: For data coming from Eclipse vectors, we can utilize https://github.com/equinor/fmu-ensemble/pull/89 when that is merged. 

For data coming from (RMS) `.csv` export, I would hope/guess it is possible to also export units at the same time as the data is exported? @HansKallekleiv?

webviz_subsurface/plugins/_inplace_volumes_onebyone.py
```diff
@@ -549,17 +603,29 @@ def calculate_table_rows(df, response):
...
+    columns = [
+        {**{"name": i[0], "id": i[0]}, **i[1]}
+        for i in InplaceVolumesOneByOne.TABLE_STATISTICS
+    ]
+    for col in columns:
+        try:
+            col["format"]["locale"]["symbol"] = [
+                "",
+                f"{VOLUME_UNITS.get(response, '')}",
+            ]
```
asnyv: for now, the "symbol" list is overwritten as long as it exists, removing the option of having anything in front of the number, could change to only overwriting the second element of the "symbol" list (which is the unit after the si-prefix) if you think it is preferable

webviz_subsurface/_abbreviations/volume_units.json
asnyv: This is basically a "sister file" to `volume_terminology.json`, so this applies mainly to the csv-files. We could also merge them, but by keeping them separate for now, I avoided having to rewrite code dependent on volume_terminology.json before we decide to go for this solution. Would of course be preferable to have units for all csv-file data as metadata from the original source, as we could then include verification of equal units when comparing data from different sources (e.g. grid data vs simulation data). The standard export files as of today are at least without specified units as far as I know.

HansKallekleiv: In Norway volumes are reported as mill Sm3 (oil) and bill Sm3 (gas). That might not be the case in other parts of the world (e.g. bbl are often used overseas). It is possible to export units from RMS, but there is no standard for this now.

I'd suggest we provide the option for the user  to overwrite keys from this file globally.

asnyv: The SI-prefix of these values are handled separately, these are only the units themselves. That is why I've written e.g. PORE_OIL in m^3, as Sm^3 doesn't make much sense for a pore volume to me. Currently dash has a bug when locking the SI prefix, but we could also include for the future a description of the prefix. E.g. that we have a single file on a form like (where prefix is optional):
{
    "RECOVERABLE_OIL": {"unit":  "Sm^3", "prefix": "M"},
    etc.
}

Another thing is that in different unit systems (like e.g. the American aka "field units" in the oil industry), the prefixes doesn't follow SI convention anyways (gas volumes in tcf -> trillion cubic feet). Allowing non-SI units will therefore open up for a whole new range of issues.

webviz_subsurface/_abbreviations/abbreviation_data/reservoir_simulation_unit_terminology.json
```diff
@@ -0,0 +1,17 @@
...
+        "RM3/DAY": "m³/day",
+        "SM3/SM3": "Sm³/Sm³",
+        "RM3/SM3": "m³/Sm³",
+        "BARSA": "bara",
+        "BARS": "barg",
+        "SM3/D/B": "Sm³/day/bar",
+        "SECONDS": "s",
+        "DAYS": "days"
+    } 
+}
```
anders-kiaer: Missing new line

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -0,0 +1,71 @@
...
+
+    return vector.split(":", 1)[0].split("_", 1)[0]
+
+
+def simulation_vector_description(vector: str):
+    """Returns a more human friendly description of the simulation vector if possible,
+     otherwise returns the input as is.
+    """
+    [vector_name, node] = vector.split(":", 1) if ":" in vector else [vector, None]
+    if len(vector_name) == 8:
```
asnyv: Found one weakness here, need to verify that metadata type is "region", other vectors may have 8 signs too, will make a fix

webviz_subsurface/_abbreviations/number_formatting.py
```diff
@@ -0,0 +1,70 @@
...
+                f"An invalid locked_si_prefix={locked_si_prefix} was used. Ignoring the locked "
+                "prefix and calculate instead."
+            ),
+            UserWarning,
+        )
+    # Calculate prefix based on value when not locked
+    # Zero is a special case
+    if number == 0:
+        return number_formatter(0, SI_PREFIXES["0"])
+    (exp_div_3, log10_rem) = divmod(math.log10(math.fabs(number)), 3)
```
anders-kiaer: Empty line before 64 perhaps.

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -0,0 +1,71 @@
...
+    otherwise returns the simulation unit.
+    * `ecl_unit`: Reservoir simulation vector unit to reformat
+    * `unit_set`: Currently only valid option is the default "METRIC" (defined as in Eclipse E100)
+    """
+    try:
+        unit = RESERVOIR_SIMULATION_UNIT_TERMINOLOGY[unit_set][ecl_unit]
+    except KeyError:
+        unit = ""
+
+    return unit
```
anders-kiaer: If we are okey with `KeyError` on unknown `unit_set`, this can be simplified to
```python
return RESERVOIR_SIMULATION_UNIT_TERMINOLOGY[unit_set].get(ecl_unit, ecl_unit)
```
In any case, shouldn't `ecl_unit` be returned and not an empty string? (If not more human friendly unit is defined).

webviz_subsurface/_private_plugins/tornado_plot.py
```diff
@@ -220,6 +233,10 @@ def _calc_tornado(reference, scale, cutbyref, data):
                     reference=reference,
                     scale=scale,
                     cutbyref=cutbyref,
+                    number_format=number_format,
+                    unit=unit,
+                    spaced=spaced,
+                    locked_si_prefix=locked_si_prefix,
```
anders-kiaer: Can we do the `data.get(...)` calls directly here instead of defining them above the `try`?

webviz_subsurface/plugins/_inplace_volumes.py
```diff
@@ -59,14 +59,68 @@ class InplaceVolumes(WebvizPluginABC):
...
+        (
+            "Maximum",
+            {
+                "type": "numeric",
+                "format": {
+                    "locale": {"symbol": ["", "unit_insert"]},
+                    "specifier": "$.4s",
+                },
+            },
+        ),
```
anders-kiaer: Use list comprehension here? As the statistics use the same dictionary.

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
```diff
@@ -68,14 +68,68 @@ class ReservoirSimulationTimeSeriesOneByOne(WebvizPluginABC):
...
+        (
+            "Maximum",
+            {
+                "type": "numeric",
+                "format": {
+                    "locale": {"symbol": ["", "unit_insert"]},
+                    "specifier": "$.4s",
+                },
+            },
+        ),
```
anders-kiaer: Since several plugins use this, should we move it to a centralized location?

webviz_subsurface/_abbreviations/reservoir_simulation.py
asnyv: I agree

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
asnyv: Is a trade-off between simplicity and configurability, but yes might be a good idea to move it. If necessary in a plugin it could be overwritten locally. Could it fit in the abbreviations/number_formatting.py?

webviz_subsurface/plugins/_inplace_volumes.py
asnyv: See next comment

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
anders-kiaer: Yes, i think that's a good place 👍

asnyv: done

PR #291, opened 2020-03-31
Show boxplot per senscase or sensname in inplace sensitivities

HansKallekleiv:


webviz_subsurface/plugins/_inplace_volumes_onebyone.py
```diff
@@ -296,7 +296,11 @@ def plot_selector(self):
                     id=self.ids("plot-type"),
```
asnyv: Have you considered to switch to a` Dropdown` to make it more compact? An extra click to switch is of course a drawback, but it is already a lot of content in the view of this plugin, so think it would be nice to keep it as clean as possible.

webviz_subsurface/plugins/_inplace_volumes_onebyone.py
```diff
@@ -451,7 +455,7 @@ def _render_vol_chart(plot_type, ensemble, response, source, *filters):
             # Make Plotly figure
             layout = {}
             layout.update(self.plotly_theme["layout"])
-            layout.update({"margin": {"l": 100}})
+            layout.update({"margin": {"l": 100, "b": 100}})
```
asnyv: Assume you are expanding the bottom margin to reduce risk of too long labels. Should we think of having a logic to expand further if the labels are very long? (Think it is nice to have a big enough margin to cover most cases by default to avoid resizing when switching plot mode like you have now)

anders-kiaer: Does [`"automargin": True`](https://plotly.com/javascript/setting-graph-size/#automatically-adjust-margins) work in this case?

asnyv: Worth a try. But if it works for the labels, it would probably result in resizing the graph when switching plot type, regardless of label length. Not a big issue, but if e.g. someone are taking snapshots of the plots to be used other places, they probably look nicer if they are on the same scale.

HansKallekleiv: yes, automargin will resize the graph when switching plot type.
Lets try with 100 margin for now ? ;)

PR #294, opened 2020-03-31
Inplace volumetrics: Initialize with boxplot if multiple ensembles are present

HansKallekleiv:
By popular demand.

### DIFF
--- a/webviz_subsurface/plugins/_inplace_volumes.py
+++ b/webviz_subsurface/plugins/_inplace_volumes.py
@@ -101,6 +101,11 @@ but the following responses are given more descriptive names automatically:
         self.uid = uuid4()
         self.selectors_id = {x: str(uuid4()) for x in self.selectors}
         self.plotly_theme = app.webviz_settings["theme"].plotly_theme
+        self.initial_plot = "Per realization"
+        self.initial_group = None
+        if len(self.volumes["ENSEMBLE"].unique()) > 1:
+            self.initial_plot = "Box plot"
+            self.initial_group = "ENSEMBLE"
         self.set_callbacks(app)
 
     def ids(self, element):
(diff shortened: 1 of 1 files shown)


webviz_subsurface/plugins/_inplace_volumes.py
```diff
@@ -101,6 +101,11 @@ def __init__(
         self.uid = uuid4()
         self.selectors_id = {x: str(uuid4()) for x in self.selectors}
         self.plotly_theme = app.webviz_settings["theme"].plotly_theme
+        self.initial_plot = "Per realization"
+        self.initial_group = None
+        if len(self.volumes["ENSEMBLE"].unique()) > 1:
+            self.initial_plot = "Box plot"
+            self.initial_group = "ENSEMBLE"
```
anders-kiaer:
```suggestion
        if len(self.volumes["ENSEMBLE"].unique()) > 1:
            self.initial_plot = "Box plot"
            self.initial_group = "ENSEMBLE"
        else:
            self.initial_plot = "Per realization"
            self.initial_group = None
```

webviz_subsurface/plugins/_inplace_volumes.py
```diff
@@ -101,6 +101,11 @@ def __init__(
         self.uid = uuid4()
         self.selectors_id = {x: str(uuid4()) for x in self.selectors}
         self.plotly_theme = app.webviz_settings["theme"].plotly_theme
+        self.initial_plot = "Per realization"
+        self.initial_group = None
+        if len(self.volumes["ENSEMBLE"].unique()) > 1:
```
asnyv: Maybe if else, rather than first setting the variables, and then overwriting them in the very likely case that there is more than one ensemble?

asnyv:
> One suggested change - otherwise LGTM (is btw. `black` not agreeing with itself?)

Aah haha, you beat me to it 🤨😅

PR #306, opened 2020-04-15
Relax version constraint dependencies, add fmu-ensemble as dependency

anders-kiaer:
Follow-up of https://github.com/equinor/fmu-ensemble/pull/130.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 with open("README.md", "r") as fh:
     LONG_DESCRIPTION = fh.read()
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -45,6 +45,7 @@
         ]
     },
     install_requires=[
+        "fmu-ensemble",
```
asnyv, line 48: How is the minimum version handled when not specified?

anders-kiaer: Then there is no minimum required version. Do we currently require what is today's latest version? I can add a version specifier.

asnyv: Will at least require the version where get_meta_data was added soon, which is pretty much the current version

PR #322, opened 2020-04-27
Support Dash 1.11

### LINKED ISSUE #317: Master failing with Dash 1.11
Suddenly the [`master` branch is failing](https://github.com/equinor/webviz-subsurface/actions?query=branch%3Amaster), due to timeout on one of the subsurface surface tests.

Investigate what has changed, and if it is needed to increase the timeout.

One way of 🔍 investigation could be to start with comparing versions used of upstream packages (use the printed CI lists of installed software from a successfull run, and a failed run, and compare).

```
    def test_surface_selector(dash_duo):
    
        app = dash.Dash(__name__)
        app.config.suppress_callback_exceptions = True
        realizations = pd.read_csv("tests/data/realizations.csv")
        s = SurfaceSelector(app, surface_context, realizations)
    
        app.layout = html.Div(children=[s.layout, html.Pre(id="pre", children="ok")])
    
        @app.callback(Output("pre", "children"), [Input(s.storage_id, "children")])
        def _test(data):
            return json.dumps(json.loads(data))
    
        dash_duo.start_server(app)
    
>       dash_duo.wait_for_contains_text("#pre", json.dumps(return_value), timeout=4)
```

HansKallekleiv:
Closes #317 

Fixes related to Dash 1.11.
- We used some wrong component props in callbacks that are no longer allowed in Dash 1.11. E.g. `dcc.Store(children)` should be `dcc.Store(data)`.
- There is a breaking change in Dash 1.11 when using `callback_context.triggered`. It is no longer possible to identify which input parameters triggers the first time a callback is executed. A check for this has been added to the necessary callbacks.

### DIFF
--- a/webviz_subsurface/_private_plugins/tornado_plot.py
+++ b/webviz_subsurface/_private_plugins/tornado_plot.py
@@ -126,7 +126,7 @@ that reads from  `tornadoplot.click_id` if `allow_click` has been specified at i
         return html.Div(
             [
                 html.Div(
-                    style={"marginLeft": "10px"},
+                    style={"marginLeft": "20%"},
                     children=[
                         html.Label(
                             "Tornado Plot",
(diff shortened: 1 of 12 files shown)


webviz_subsurface/_private_plugins/tornado_plot.py
```diff
@@ -126,7 +126,7 @@ def layout(self):
         return html.Div(
             [
                 html.Div(
-                    style={"marginLeft": "10px"},
+                    style={"marginLeft": "20%"},
```
anders-kiaer, line 129: This change perhaps belong in another PR?

Issue #296, opened 2020-04-01
Improvement suggestion to in-place volumes table


tnatt:
- [ ] For readability move units to table header

- [ ] Reduce number of decimals in std.dev

- [ ] Option to remove min and max from table

asnyv:
Had a chat with @tnatt regarding this issue, as some of the required modifications in [ReservoirSimulationTimeseriesOneByOne](https://github.com/equinor/webviz-subsurface/blob/master/webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py) overlaps with code that has to be modified in #284 (metadata for time series).

Summary of chat:
- Units to table header: Make a title for each table with the Response/Time series and unit, so that the tables consist of all the basic information and can be copied to e.g. a report. (Most plots already have titles and/or axes). Also means that the content of the `Response` column in `InplaceVolumes.py` (which is a constant value for all rows) can be moved to the title instead.
- Should reduce number of digits by 1 in stddev as it is usually a lower value than the other columns, therefore (maybe) more likely to fit the number of decimals with one digit less.

PR #323, opened 2020-04-27
Simulation metadata

### LINKED ISSUE #284: Utilize metadata from fmu-ensemble
Utilize https://github.com/equinor/fmu-ensemble/pull/89 in `webviz-subsurface`.

E.g.:
- more robust detection of historical vs. non-historical vectors
- more robust split of vector name vs. node name.
- show human-readable y-axis labels
- time series interpolation method based on vector type (cumulative vector linear interpolation, rate vector backfilled). See [`linear` and `vh` interpolation in `plotly`](https://plotly.com/python/line-charts/#interpolation-with-line-plots).

asnyv:
Resolves #284. 

The "more robust vector vs node name split" was not done, as it seems like `fmu_ensemble`/`libecl` doesn't support all the properties of vector names that we use (especially for region keywords). Think our existing solution is the best we could hope for atm.

Some improvements have also been made to the vector type identification (`is_rate`/`is_total`) [upstream in libecl](https://github.com/equinor/libecl/pull/707).  but the code is not dependent on these changes. 
Will also improve the `is_historical` flag when [this issue is completed ](https://github.com/equinor/libecl/issues/708). Should be pretty straight forward as long as we get the ideas "approved".

Since a lot of the modifications were related to vectors and units, I tried to keep #296 in the back of my head. Point 1 and 2 in #296 is therefore solved in this PR.

Also extended list of "human readable units" from the Eclipse `METRIC` unit format

### DIFF
--- a/webviz_subsurface/_abbreviations/reservoir_simulation.py
+++ b/webviz_subsurface/_abbreviations/reservoir_simulation.py
@@ -1,6 +1,9 @@
 import json
 import pathlib
 import warnings
+from typing import Optional
+
+import pandas as pd
 
 
 _DATA_PATH = pathlib.Path(__file__).parent.absolute() / "abbreviation_data"
@@ -14,7 +17,7 @@ RESERVOIR_SIMULATION_UNIT_TERMINOLOGY = json.loads(
 )
 
 
-def simulation_unit_reformat(ecl_unit: str, unit_set: str = "METRIC"):
+def simulation_unit_reformat(ecl_unit: str, unit_set: str = "METRIC") -> str:
     """Returns the simulation unit in a different, more human friendly, format if possible,
     otherwise returns the simulation unit.
     * `ecl_unit`: Reservoir simulation vector unit to reformat
@@ -23,15 +26,14 @@ def simulation_unit_reformat(ecl_unit: str, unit_set: str = "METRIC"):
     return RESERVOIR_SIMULATION_UNIT_TERMINOLOGY[unit_set].get(ecl_unit, ecl_unit)
 
 
-def simulation_vector_base(vector: str):
+def simulation_vector_base(vector: str) -> str:
     """Returns base name of simulation vector
     E.g. WOPR for WOPR:OP_1 and ROIP for ROIP_REG:1
     """
-
     return vector.split(":", 1)[0].split("_", 1)[0]
 
 
-def simulation_vector_description(vector: str):
+def simulation_vector_description(vector: str) -> str:
     """Returns a more human friendly description of the simulation vector if possible,
      otherwise returns the input as is.
     """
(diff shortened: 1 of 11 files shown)


webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -71,3 +73,41 @@ def simulation_vector_description(vector: str):
         )
 
     return description
+
+
+def historical_vector(
+    vector: str,
+    smry_meta: Optional[pd.DataFrame] = None,
```
anders-kiaer, line 80: Maybe add some docstring explanation of the `smry_meta` input, and how that is used in this function.

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
```diff
@@ -336,13 +388,18 @@ def _render_date(ensemble, clickdata, vector):
                 # json.dumps(f"{date}"),
                 table_rows,
                 table_columns,
+                (
+                    f"{simulation_vector_description(vector)} ({vector})"
+                    if get_unit(self.smry_meta, vector) == ""
+                    else f"{simulation_vector_description(vector)} ({vector})"
+                    + f" [{get_unit(self.smry_meta, vector)}]"
+                ),
```
anders-kiaer, line 396:
```suggestion
                f"{simulation_vector_description(vector)} ({vector})" + ("" if get_unit(self.smry_meta, vector) == "" else f" [{get_unit(self.smry_meta, vector)}]")
```

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
```diff
@@ -476,3 +550,8 @@ def filter_ensemble(data, ensemble, vector):
 @webvizstore
 def read_csv(csv_file) -> pd.DataFrame:
     return pd.read_csv(csv_file, index_col=None)
+
+
+@CACHE.memoize(timeout=CACHE.TIMEOUT)
+def get_unit(smry_meta, vec):
+    return "" if smry_meta is None else simulation_unit_reformat(smry_meta.unit[vec])
```
anders-kiaer, line 557: Would it be simpler/better that this returns `None` when there is no unit, instead of empty string?

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -129,13 +120,23 @@ def __init__(
...
+        line_shape_fallback = line_shape_fallback.lower()
+        if line_shape_fallback in ("backfilled", "backfill"):
+            self.line_shape_fallback = "vh"
+        elif line_shape_fallback in ["hv", "vh", "hvh", "vhv", "spline", "linear"]:
+            self.line_shape_fallback = line_shape_fallback
+        else:
+            self.line_shape_fallback = "linear"
+            warnings.warn(
+                f"{line_shape_fallback} is not a valid line_shape_fallback option, will use linear."
+            )
```
anders-kiaer, line 139: Similar code blocks in multiple plugins. Move to common function?

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
asnyv: It is probably "better", and was considering it, but empty string is "safer" in the sense that no errors are thrown if you don't include `None` handling each place the `get_unit` is used. Quite easy to change it still, so I can if you want me to.

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
asnyv: Could do that, not sure where it would naturally belong though? Doesn't quite fit neither `datainput` or `abbreviations` as I think of those terms.

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
asnyv: 👍  will change this dependent on the conclusion on `get_unit()`

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
anders-kiaer: Vote for changing to `None`. Then I guess it would also be easy to distinguish between missing/undefined unit, and a unit-less quantity/ratio?

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
anders-kiaer: Not sure either. A new folder `_utils` containing small utility functions used by plugins not covered by the two existing subfolders?

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
asnyv: 👍 will do. And yes, that should be easy.

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
asnyv: Think it is the cleanest option at least.

asnyv:
https://github.com/equinor/libecl/pull/707 merged upstream

Issue #321, opened 2020-04-27
Eclipse UDQs


havarhal:
When loading webviz, warnings related to Eclipse user define quantities(UDQs) appear. Suggest to turn off the check for descriptions of vectors starting with 'CU',' FU','GU','RU','SU','WU', 'AU' and 'BU'.

asnyv:
Good point @havarhal 👍 Would it be useful to add a description like "VECTOR (UDQ)" to these vectors? To make it possible to search for UDQs?

havarhal:
You mean like a user defined description?

asnyv:
Was thinking just a general. More like a "UDQ" tag behind the vector name. Have a feeling that most users will not bother to make user defined ones even if that was an option?

havarhal:
I agree, but adding UDQ as a general description was not a bad idea :)

anders-kiaer:
>Have a feeling that most users will not bother to make user defined ones even if that was an option?

Many users might not do that effort, but perhaps we should add the possibility. I know that some assets want to make time series more available to non-reservoir-engineers, and giving the user the possibility to provide some custom descriptions of vectors could be useful (especially for e.g. UDQ vectors which might be quite difficult to interpret for persons not setting up the simulation model).

This "standard customization" can be added to e.g. let the user provide metadata in the form of `(time, vector, some text)` in order to annotate given plots (where then e.g. a vertical line with some text appears at the given time. Could be useful for curves where the user wants to tell a "story" (e.g. tracer curves, annotate when the tracer component was injected, explain some sharp changes in rate curves etc.).

Could be as simple as an optional `yaml` input file to the `ReservoirTimeSeriesPlugin`, where vector descriptions are merged with the standard provided dictionary.

asnyv:
Think you should then split the UDQ description and plot annotations in separate structures. Too many options in the same file are confusing, especially when it comes to what is optional, and what is required. 

The UDQ description is also something which is "global" in the app (at least for an ensemble, can probably also assume it is for an app). Many plugins have simulation vectors, like `ReservoirSimulationTimeSeries`, `ReservoirSimulationTimeSeriesOneByOne`, `ParameterResponseCorrelation`, future plugin for simulated inplace and recovery, and probably several AHM plugins and etc in the future. Shouldn't have to define this file for each plugin.
Annotations are on the other hand more for specific plots in my opinion (though they could of course be stored "globally").

asnyv:
[...]

PR #332, opened 2020-04-30
Improved vector description for completions and UDQs

asnyv:
1. Deactivates the description warnings for UDQs as originally requested in #321
  -Does not implement the additional suggestions on user defined UDQ descriptions.
2. Fixes issue with well completion keyword descriptions (e.g. WOPRL) that was discovered during testing of point 1.

### DIFF
--- a/webviz_subsurface/_abbreviations/reservoir_simulation.py
+++ b/webviz_subsurface/_abbreviations/reservoir_simulation.py
@@ -30,7 +30,7 @@ def simulation_vector_base(vector: str) -> str:
     """Returns base name of simulation vector
     E.g. WOPR for WOPR:OP_1 and ROIP for ROIP_REG:1
     """
-    return vector.split(":", 1)[0].split("_", 1)[0]
+    return vector.split(":", 1)[0].split("_")[0]
 
 
 def simulation_vector_description(vector: str) -> str:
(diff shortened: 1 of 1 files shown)


webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -30,47 +30,62 @@ def simulation_vector_base(vector: str) -> str:
...
+            # 8 signs, with the last 3 defining the region.
+            # E.g.: For an array "FIPREG": ROIP is ROIP_REG, RPR is RPR__REG and ROIPL is ROIPLREG
+            # Underscores _ are always used to fill
+            [vector_base_name, fip] = [vector_name[0:5].rstrip("_"), vector_name[5:]]
+            try:
+                if SIMULATION_VECTOR_TERMINOLOGY[vector_base_name]["type"] == "region":
+                    vector_name = vector_base_name
+            except KeyError:
+                pass
+        elif vector_name.startswith(("W")) and vector_name[4] == "L":
```
anders-kiaer, line 55:
```suggestion
        elif vector_name.startswith("W") and vector_name[4] == "L":
```

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -30,47 +30,62 @@ def simulation_vector_base(vector: str) -> str:
...
-        description = vector_name
-        warnings.warn(
-            (
-                f"Could not find description for vector {vector_name}. Consider adding"
-                " it in the GitHub repo https://github.com/equinor/webviz-subsurface?"
-            ),
-            UserWarning,
-        )
+        description = vector
+        if not vector.startswith(("AU", "BU", "CU", "FU", "GU", "RU", "SU", "WU")):
```
anders-kiaer, line 81: Add a comment why we don't give warning on these vectors, for those not into the current issue.

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -30,47 +30,62 @@ def simulation_vector_base(vector: str) -> str:
...
+            # Region vectors for other FIP regions than FIPNUM are written on a special form:
+            # 8 signs, with the last 3 defining the region.
+            # E.g.: For an array "FIPREG": ROIP is ROIP_REG, RPR is RPR__REG and ROIPL is ROIPLREG
+            # Underscores _ are always used to fill
+            [vector_base_name, fip] = [vector_name[0:5].rstrip("_"), vector_name[5:]]
+            try:
+                if SIMULATION_VECTOR_TERMINOLOGY[vector_base_name]["type"] == "region":
+                    vector_name = vector_base_name
+            except KeyError:
+                pass
```
anders-kiaer, line 54: `type` should always be defined I guess, if `vector_base_name` exists, so catching `KeyError` might be slightly loose? Maybe
```suggestion
            if vector_base_name in SIMULATION_VECTOR_TERMINOLOGY and SIMULATION_VECTOR_TERMINOLOGY[vector_base_name]["type"] == "region":
                    vector_name = vector_base_name
```

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -30,47 +30,62 @@ def simulation_vector_base(vector: str) -> str:
...
         if node is not None:
             description += (
-                f", {metadata['type'].replace('_', ' ')} {fip} {node}"
-                if fip is not None
-                else f", {metadata['type'].replace('_', ' ')} {node}"
+                f", {metadata['type'].replace('_', ' ')} {node}"
+                if fip is None and comp is None
+                else f", {metadata['type'].replace('_', ' ')} {fip} {node}"
+                if comp is None
+                else f", well {node} completion {comp}"
```
anders-kiaer, line 77: This starts to become complex to read. Maybe we could utilize that in earlier `if/else` branches of this function, we know when we have `fip` or `comp` defined, and instead do early returns?

webviz_subsurface/_abbreviations/reservoir_simulation.py
asnyv: Would say that these two options are equally loose imo, but can change it. Should do it for `type == completion` too in that case.

webviz_subsurface/_abbreviations/reservoir_simulation.py
asnyv: We could. Didn't shuffle more code than I had too for this PR, but I agree that it is getting a bit messy.

webviz_subsurface/_abbreviations/reservoir_simulation.py
anders-kiaer: The first will fail the `try` clause both if `vector_base_name` is not defined, and if `type` is (wrongly?) missing from the definition (if `vector_base_name` exists). The latter will not silently fail if the definition is (wrongly?) missing a `type` key for some vector that exists.

Yes, if we change we should change the other branches as well.

asnyv: I agree that from our side it is best that it fails, but for a user it might be better that it is quite ;) 
But of course, if we later include the option of user defined extensions, then an error might be a good thing.
I'll change it.

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -39,38 +39,53 @@ def simulation_vector_description(vector: str) -> str:
...
     if vector_name in SIMULATION_VECTOR_TERMINOLOGY:
         metadata = SIMULATION_VECTOR_TERMINOLOGY[vector_name]
         description = metadata["description"]
         if node is not None:
-            description += (
-                f", {metadata['type'].replace('_', ' ')} {fip} {node}"
-                if fip is not None
-                else f", {metadata['type'].replace('_', ' ')} {node}"
-            )
+            description += f", {metadata['type'].replace('_', ' ')} {node}"
```
anders-kiaer, line 75: Could `return` here as well.

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -39,38 +39,53 @@ def simulation_vector_description(vector: str) -> str:
...
+            # https://github.com/equinor/webviz-subsurface/issues/321)
+            warnings.warn(
+                (
+                    f"Could not find description for vector {vector_name}. Consider adding"
+                    " it in the GitHub repo https://github.com/equinor/webviz-subsurface?"
+                ),
+                UserWarning,
+            )
 
     return description
```
anders-kiaer, line 90: And then here instead `return vector`?

PR #336, opened 2020-05-02
Make all plugin settings persistent

HansKallekleiv:
Add [persistence](https://dash.plotly.com/persistence) to all components with user settings.
Relates to https://github.com/equinor/webviz-config/issues/223


HansKallekleiv:
Replaced by #347

PR #337, opened 2020-05-03
RFT Viewer

tnatt:


asnyv:
Will give a more thorough code review later, this is mainly based on the current look of it in the demo (and a short look at the code):

In short it looks nice :) But at least what I call the "biggest issue" below should be looked at.

**RFT Map**
Just minor things:
- Would be nice to know if it is the maximum (absolute) misfit, the mean misfit or a normalized (the last is clearly not the case based on the color scale) when you color/size by misfit.
- As only TVD is currently supported, you could add that to the "Depth" label on the Y-axis of the first page.

**RFT misfit per real**
Biggest issue in my opinion:
- Hard to understand how many RFTs/data points that are actually part of this view. I don't know if the same well is added several times (due to multiple RFTs), if a well is missing, if the wells exist at all dates and etc without playing with the filters to sort out the possible combinations. A cumulative value doesn't add much value if you don't know what you have aggregated. Might be better to have just ensembles and observations as filters to make it clear what you are looking at?
Alternatively that the currently selected observations based on the filters are listed somewhere in the view.

**RFT misfit per real**
Minor things:
- The current filters makes more sense here, since each observation gets its own data point in the cross-plot.
- Would be interesting to be able to change y-axis between simulated mean, simulated max and simulated min in the ensemble.

.github/workflows/webviz-subsurface.yml
```diff
@@ -74,10 +74,10 @@ jobs:
       env:
         # If you want the CI to (temporarily) run against your fork of the testdada,
         # change the value her from "equinor" to your username.
-        TESTDATA_REPO_OWNER: equinor
+        TESTDATA_REPO_OWNER: hanskallekleiv
         # If you want the CI to (temporarily) run against another branch than master,
         # change the value her from "master" to the relevant branch name.
-        TESTDATA_REPO_BRANCH: master
+        TESTDATA_REPO_BRANCH: rft-data
```
anders-kiaer, line 80: Change back.

webviz_subsurface/plugins/_rft_plotter/_misfit_figure.py
```diff
@@ -0,0 +1,102 @@
+import webviz_core_components as wcc
+
+
+def enscolors():
```
anders-kiaer, line 4: Could/should this instead be a static constant list `ENSCOLORS`?

webviz_subsurface/plugins/_rft_plotter/rft_plotter.py
```diff
@@ -0,0 +1,754 @@
...
+        if csvfile_rft and ensembles:
+            raise ValueError(
+                'Incorrent arguments. Either provide a "csvfile_rft" and "csvfile_rft_ert" or '
+                '"ensembles"'
+            )
+        if csvfile_rft and csvfile_rft_ert:
+            self.simdf = read_csv(self.csvfile_rft)
+            self.ertdatadf = read_csv(self.csvfile_rft_ert)
+
+        elif ensembles:
```
anders-kiaer, line 123:
```suggestion
            )

        if csvfile_rft and csvfile_rft_ert:
            self.simdf = read_csv(self.csvfile_rft)
            self.ertdatadf = read_csv(self.csvfile_rft_ert)
        elif ensembles:
```

webviz_subsurface/plugins/_rft_plotter/rft_plotter.py
```diff
@@ -0,0 +1,754 @@
...
+                else None
+            )
+
+            try:
+                self.simdf = load_csv(self.ens_paths, "share/results/tables/rft.csv")
+            except KeyError:
+                raise KeyError(
+                    "Csv file for Eclipse RFT output (share/results/tables/rft.csv) not found!"
+                )
+            try:
```
anders-kiaer, line 141:
```suggestion
                )

            try:
```

webviz_subsurface/plugins/__init__.py
```diff
@@ -45,7 +45,7 @@
 from ._parameter_parallel_coordinates import ParameterParallelCoordinates
 from ._running_time_analysis_fmu import RunningTimeAnalysisFMU
 from ._relative_permeability import RelativePermeability
-
+from ._rft_plotter.rft_plotter import RftPlotter
```
anders-kiaer, line 48:
```suggestion
from ._rft_plotter.rft_plotter import RftPlotter

```

setup.py
```diff
@@ -42,9 +42,11 @@
             + "webviz_subsurface.plugins:ParameterParallelCoordinates",
             "RunningTimeAnalysisFMU = webviz_subsurface.plugins:RunningTimeAnalysisFMU",
             "RelativePermeability = webviz_subsurface.plugins:RelativePermeability",
+            "RftPlotter =  webviz_subsurface.plugins:RftPlotter",
         ]
     },
     install_requires=[
+        "colorlover>=0.3.0",
```
anders-kiaer, line 49: Is it possible to use https://plotly.com/python/builtin-colorscales/ instead of adding `colorlover` as dependency?

anders-kiaer: Current `colorlover` interpolation:
`hsl(210.0, 100.0%, 98.0%)`
`hsl(208.0, 60.0%, 91.0%)`
`hsl(209.0, 56.0%, 85.0%)`
`hsl(200.0, 52.0%, 75.0%)`
`hsl(202.0, 56.0%, 62.0%)`

`f`-string, with endpoints `Blues[0]` and `Blues[5]`:
`hsl(210.0, 100.0%, 98.0%)`
`hsl(208.0, 89.0%, 89.0%)`
`hsl(206.0, 78.0%, 80.0%)`
`hsl(204.0, 67.0%, 71.0%)`
`hsl(202.0, 56.0%, 62.0%)`

`f`-string*, with endpoints `Blues[0]` and `Blues[6]`:
`hsl(210.0, 100.0%, 98.0%)`
`hsl(208.25, 88.25%, 86.25%)`
`hsl(206.5, 76.5%, 74.5%)`
`hsl(204.75, 64.75%, 62.75%)`
`hsl(203.0, 53.0%, 51.0%)`

`f`-string, with endpoints `Blues[0]` and `Blues[-1]`:
`hsl(210.0, 100.0%, 98.0%)`
`hsl(210.75, 97.25%, 81.0%)`
`hsl(211.5, 94.5%, 64.0%)`
`hsl(212.25, 91.75%, 47.0%)`
`hsl(213.0, 89.0%, 30.0%)`

\* Corresponding `f`-string:
```python
[f"hsl({210*(1-x) + 203*x}, {100*(1-x) + 53*x}%, {98*(1-x)+51*x}%)" for x in np.linspace(0, 1, 5)]
```

PR #343, opened 2020-05-06
Correct realizations in TornadoPlot tooltip

### LINKED ISSUE #340: TornadoPlot realizations in tooltips incorrect
The realization intervals in tooltips are only defined as the range between the max and min real number, incorrect if the realizations for the low or high case in the `TornadoPlot` are not a list of successive integers.

asnyv:
Resolves #340

### DIFF
--- a/webviz_subsurface/_private_plugins/tornado_plot.py
+++ b/webviz_subsurface/_private_plugins/tornado_plot.py
@@ -1,7 +1,8 @@
 from uuid import uuid4
 import json
-import pandas as pd
+from typing import List, Optional
 
+import pandas as pd
 import dash
 from dash.dependencies import Input, Output
 from dash.exceptions import PreventUpdate
@@ -355,10 +356,10 @@ def tornado_plot(
             p10 = case_df["VALUE"].quantile(0.90)
 
             # Extract list of realizations with values less then reference avg (low)
-            low_reals = list(case_df.loc[case_df["VALUE"] <= ref_avg]["REAL"])
+            low_reals = list(map(int, case_df.loc[case_df["VALUE"] <= ref_avg]["REAL"]))
 
             # Extract list of realizations with values higher then reference avg (high)
-            high_reals = list(case_df.loc[case_df["VALUE"] > ref_avg]["REAL"])
+            high_reals = list(map(int, case_df.loc[case_df["VALUE"] > ref_avg]["REAL"]))
 
             arr.append(
                 {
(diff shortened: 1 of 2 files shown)


asnyv:
Will update commit message and reinsert the blank line from the workflow after review.

webviz_subsurface/_private_plugins/tornado_plot.py
```diff
@@ -574,3 +575,27 @@ def calc_low_x(low, high):
...
+            if int_list[i] != (int_list[i - 1] + 1):
+                if i == 1:
+                    return f"{int_list[0]}, {_printable_int_list_rec(int_list[i:])}"
+                return f"{int_list[0]}-{int_list[i-1]}, {_printable_int_list_rec(int_list[i:])}"
+        return f"{int_list[0]}-{int_list[-1]}"
+
+    if integer_list is not None and len(integer_list) > 0:
+        integer_list = sorted(integer_list)
+        return _printable_int_list_rec(integer_list)
+    return "None"
```
anders-kiaer, line 601:
```suggestion
    if not integer_list:
        return "None"

    sorted_list = sorted(integer_list)
    prev_number = sorted_list[0]
    string = str(prev_number)

    for real in sorted_list[1:]:
        if real > prev_number + 1:
            string += (f"{prev_number}" if string.endswith("-") else "") + f", {real}"
        elif not string.endswith("-"):
            string += "-"
        prev_number = real

    if not string.endswith(f", {prev_number}"):
        string += str(prev_number)

    return string
```
Could this be an alternative non-recursive way? Four more lines, but perhaps more readable?

webviz_subsurface/plugins/_reservoir_simulation_timeseries_onebyone.py
```diff
@@ -206,7 +206,7 @@ def ensemble_selector(self):
             style={"paddingBottom": "30px"},
             children=html.Label(
                 children=[
-                    html.Span("Ensemble", style={"font-weight": "bold"}),
+                    html.Span("Ensemble:", style={"font-weight": "bold"}),
```
anders-kiaer, line 209: Not in scope of this PR?

asnyv: No, leftover from the "Clear change" I tested yesterday for the deploy test. Will move it to the PR on fixing the onebyone.

webviz_subsurface/_private_plugins/tornado_plot.py
asnyv: Agree that it is probably easier to follow. Tested performance, and they are pretty much the same (the fastest varied a bit dependent on input). Will change the variable `real` to `next_number` though, as I find it best to keep it generic in case we need a similar method for something else later. Also, this suggestion can't handle lists of length 1, but that is a simple fix of course

asnyv: Also added a test for `len(integer_list)==1`

PR #348, opened 2020-05-28
Regional time series plugin

### LINKED ISSUE #278: Dynamic data in inplace volumes plugin
Include Eclipse summary data in Inplace Volumes plugin. Should support (both for Field F and region R, but only writing F here):
In-place: FOIP, FGIP, FWIP
   Potentially also more detailed: like FGIPL, FGIPG (gas in place in liquid and gas phases), FOIPL/FOIPG Recovery: FOE exists for oil, but should support calculated from FOIP and FGIP.
FHPV (hydrocarbon pore volume), FOPV/FGPV/FWPV (pore volume containting oil/gas/water)
FRPV (reservoir condition pore volume

Should also allow other user defined FIP regions than FIPNUM (propose to use FIPXXX as name of source in the source dropdown). 

Comparison with csv-files allowed only for initial in place. Have a switch button between initial and dynamic to clearly separate? Dynamic activates a date dropdown and removes all static data.

Need a config file for FIPXXX regions to e.g. Zone and Region:
- FIPNUM:
  - REGION:
     - 1: [1, 3, 5]
     - 2: [2, 4, 6]
  - ZONE:
     - UpperReek: [1, 2]
     - MidReek: [3, 4]
     - LowerReek: [5, 6]
 - (Any other definition the user wants), these will be added to filters

As the user in the csv-file data can decide which data they want to include in the different columns (e.g. if GIIP_GAS with or without solution gas/associated gas and etc), it is a bit risky to make assumptions on the matching between csv-file columns and Eclipse summary vectors. Should therefore require a mapping by the user if comparison with Eclipse data shall be allowed.
Can be combined in a "volumes_config.yaml" with e.g.
fip:
  - FIPNUM
    - xxxx
  - FIPXXX
    - xxxx

csv_eclipse_map:
  - STOIIP_OIL: [FOIP, ROIP]
  - HCPV_OIL: [FOPV, ROPV]
  and etc

asnyv:
Resolves #278 
New plugin aggregates and calculates recovery for regional simulation time series vectors on the fly.  Supported vectors: In-place, rates and cumulatives for oil, gas and water + recovery of oil and gas.
An optional yaml file can be defined to specify your own filters instead of just the Eclipse region numbers.

Will replace #287 with this one, as it was decided not to mix in time series data in the `InplaceVolumes` plugin.


webviz_subsurface/plugins/_reservoir_simulation_timeseries_regional.py
```diff
@@ -0,0 +1,1305 @@
...
+        ]
+
+    def set_callbacks(self, app):
+        @app.callback(
+            [
+                Output(self.uuid("graph"), "figure"),
+                Output(self.uuid("date_view_wrapper"), "children"),
+                Output(self.uuid("ref_vec"), "data"),
+            ],
+            self._callback_inputs + [Input(self.uuid("date"), "data")],
```
HansKallekleiv, line 601: This might be a good opportunity to look at pattern matched callbacks https://dash.plotly.com/pattern-matching-callbacks

The select id's could be replaced by a dictionary and then pattern matching can be used in the callback.
```python
    def selectors_id(self, selected):
        return {
            "page": self.uuid(f"select"),
            "type": "select",
            "value": selected,
        }

def set_callbacks(self, app):
        @app.callback(
            [
                Output(self.uuid("graph"), "figure"),
                Output(self.uuid("date_view_wrapper"), "children"),
                Output(self.uuid("ref_vec"), "data"),
            ],
            [
                Input(
                    {
                        "page": self.uuid(f"select"),
                        "type": "select",
                        "value": ALL,
                    }, 'value'
                )
            ]
            + [Input(self.uuid("date"), "data")]
            
        )
        def _render_charts(*args):  # pylint: disable=too-many-locals
            inputs = dash.callback_context.inputs
            #...
```

asnyv: @HansKallekleiv  yes it could be a good idea. 
I've tested a bit more, and I think we have to make individual sets of selectors/filters for each FIP array selection, as they don't necessarilly have the same filters (e.g. not all are geolocial zone/region, but maybe FIPOWC is above/below initial contact and etc). Could then maybe set the type to the value of the FIP array selector, so that only the relevant filters are active. Should ideally hide the inactive ones as well. In theory we could probably with the pattern matching callbacks even create new filters when switching FIP array instead of hiding/unhiding?

HansKallekleiv: yes should be possible :)

.github/workflows/webviz-subsurface.yml
```diff
@@ -74,10 +74,10 @@ jobs:
       env:
         # If you want the CI to (temporarily) run against your fork of the testdada,
         # change the value her from "equinor" to your username.
-        TESTDATA_REPO_OWNER: equinor
+        TESTDATA_REPO_OWNER: asnyv
         # If you want the CI to (temporarily) run against another branch than master,
         # change the value her from "master" to the relevant branch name.
-        TESTDATA_REPO_BRANCH: master
+        TESTDATA_REPO_BRANCH: regional
```
anders-kiaer, line 80: Change back

setup.py
```diff
@@ -42,10 +42,12 @@
             + "webviz_subsurface.plugins:ParameterParallelCoordinates",
             "RunningTimeAnalysisFMU = webviz_subsurface.plugins:RunningTimeAnalysisFMU",
             "RelativePermeability = webviz_subsurface.plugins:RelativePermeability",
+            "ReservoirSimulationTimeSeriesRegional = "
+            + "webviz_subsurface.plugins:ReservoirSimulationTimeSeriesRegional",
         ]
     },
     install_requires=[
-        "fmu-ensemble>=1.2.2",
+        "fmu-ensemble>=1.2.3",
```
anders-kiaer, line 50: Is this a required bump wrt. this PR?

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -30,7 +30,7 @@ def simulation_vector_base(vector: str) -> str:
     """Returns base name of simulation vector
     E.g. WOPR for WOPR:OP_1 and ROIP for ROIP_REG:1
     """
-    return vector.split(":", 1)[0].split("_")[0]
+    return vector.split(":", 1)[0].split("_", 1)[0][:5] if ":" in vector else vector
```
anders-kiaer, line 33: Despite being a one-liner, the logic is a bit hard to read for newcomers. Maybe add an example in the docstring of why e.g. `[:5]` is needed?

webviz_subsurface/_abbreviations/reservoir_simulation.py
```diff
@@ -49,7 +49,6 @@ def simulation_vector_description(vector: str) -> str:
                 vector_base_name in SIMULATION_VECTOR_TERMINOLOGY
                 and SIMULATION_VECTOR_TERMINOLOGY[vector_base_name]["type"] == "region"
             ):
-                print(1)
```
anders-kiaer, line 52: Remove

webviz_subsurface/plugins/_reservoir_simulation_timeseries_regional.py
```diff
@@ -0,0 +1,1351 @@
...
+            },
+            {
+                "id": self.uuid("fip_array"),
+                "content": (
+                    "Select the Eclipse FIP array, e.g. FIPNUM or a custom FIPXXX"
+                ),
+            },
+            {
+                "id": self.uuid("groupby"),
+                "content": ("Select how the data should be grouped. Ensemble"),
```
anders-kiaer, line 201: Remove ( )?

setup.py
asnyv: `1.2.3` is the first version after https://github.com/equinor/fmu-ensemble/pull/134 where the`first` option used for `FOPT` `FGPT` and `FWPT` to check if the ensemble lacks history was introduced. Possible to bypass it by loading e.g. yearly instead and filter later on if you want to. `first` is simpler and stores less data though.

PR #364, opened 2020-06-19
ParameterResponseCorrelation responses from UNSMRY

### LINKED ISSUE #346: Utilize fmu-ensemble for responses in ParameterResponseCorrelation
`ParameterResponseCorrelation` currently only support csv-input for responses, even though parameters can be read by the use of `fmu-ensemble`. Should add option to load summary data through `fmu-ensemble`. Would also allow meta data with e.g. units for the response.

asnyv:
Resolves #346

Additionally: Empty "distribution graph" (the scatter and historgram plots) instead of `PreventUpdate` when changing to a filter combination that doesn't work, similar to the "correlation graph" (the tornado plot)

### DIFF
--- a/webviz_subsurface/plugins/_parameter_response_correlation.py
+++ b/webviz_subsurface/plugins/_parameter_response_correlation.py
@@ -4,7 +4,6 @@ from pathlib import Path
 import numpy as np
 import pandas as pd
 from plotly.subplots import make_subplots
-from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output
 import dash_html_components as html
 import dash_core_components as dcc
@@ -14,7 +13,7 @@ from webviz_config.common_cache import CACHE
 from webviz_config import WebvizPluginABC
 from webviz_config.utils import calculate_slider_step
 
-from .._datainput.fmu_input import load_parameters, load_csv
+from .._datainput.fmu_input import load_parameters, load_csv, load_smry
 
 
 class ParameterResponseCorrelation(WebvizPluginABC):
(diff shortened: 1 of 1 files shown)


.github/workflows/webviz-subsurface.yml
```diff
@@ -74,10 +74,10 @@ jobs:
       env:
         # If you want the CI to (temporarily) run against your fork of the testdada,
         # change the value her from "equinor" to your username.
-        TESTDATA_REPO_OWNER: equinor
+        TESTDATA_REPO_OWNER: asnyv
         # If you want the CI to (temporarily) run against another branch than master,
         # change the value her from "master" to the relevant branch name.
-        TESTDATA_REPO_BRANCH: master
+        TESTDATA_REPO_BRANCH: paramrespcorrfmuensemble
```
anders-kiaer, line 80: Change back.

webviz_subsurface/plugins/_parameter_response_correlation.py
```diff
@@ -36,14 +35,21 @@ class ParameterResponseCorrelation(WebvizPluginABC):
...
+               the date will be used as a response filter of type `single`.
 * `response_file`: Local (per realization) csv file for response parameters.
 * `response_filters`: Optional dictionary of responses (columns in csv file) that can be used
 as row filtering before aggregation. (See below for filter types).
-
 * `response_ignore`: Response (columns in csv) to ignore (cannot use with response_include).
 * `response_include`: Response (columns in csv) to include (cannot use with response_ignore).
-* `aggreation`: How to aggregate responses per realization. Either `sum` or `mean`.
+* `column_keys`: Simulation vectors to use as responses read directly from UNSMRY-files in the
+                 defined ensembles using fmu-ensemble. (cannot use with response_file,
```
anders-kiaer, line 48: Remove unnecessary .

asnyv: Not following you? The white line on line 43 or?

PR #365, opened 2020-06-24
TornadoPlot modifications from user feedback

asnyv:
- [x] Fix bug in TornadoPlot tooltips when both cases are on the same side of reference.

- [x] Support single case sensitivities

- [x] Deactivate SI prefixed numbers when using percentage. 

Note that I also refactored calculation of the DataFrame out of the `tornado_plot` method, as the `tornado_plot` method was beginning to get quite long.

### DIFF
--- a/webviz_subsurface/_private_plugins/tornado_plot.py
+++ b/webviz_subsurface/_private_plugins/tornado_plot.py
@@ -330,8 +330,135 @@ def tornado_plot(
         )
     ]["VALUE"].mean()
 
+    df = calc_tornado_df(data, realizations, ref_avg, scale)
+
+    # Drops sensitivities smaller than reference if specified
+    if cutbyref and df["sensname"].str.contains(reference).any():
+        df = cut_by_ref(df, reference)
+
+    df = sort_by_max(df)
+
+    store_low_high = {
+        sensname: {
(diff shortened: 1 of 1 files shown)


asnyv:
Note that I also refactored calculation of the DataFrame out of the `tornado_plot` method as it was beginning to get quite long.

Think that it behaves now as requested by user for the single cohiba sensitivity in the demo, but I believe myself that it would be better to handle the cohiba sensitivity as a "Monte Carlo" type sensitivity instead of "scalar", as the sensitivity as it is now seems much smaller in the `TornadoPlot` than it is in reality. Sensitivity type is though not handled by `TornadoPlot` itself, so would be a separate PR in that case. Also, the change would be a bit breaking as casename "simulation" would have to be added to "MC" style sensitivities (unless we would say that _**all**_ single case sensitivities should be handled as "MC", but  I believe that could have unfortunate consequences if e.g. users run single realizations per sensitivity.

asnyv:
Regarding comment on Cohiba sensitivity:
User agreed that the current Cohiba case would be better off with a MC style sensitivity, but that this is an input issue, and that the code in this PR gives the expected behavior for a single case sensitivity.

PR #375, opened 2020-07-03
Update DiskUsage to newer standard file location and format.

### LINKED ISSUE #374: DiskUsage no longer compatible with standard
A while back the standard location for scratch-disc usage was moved to a folder `.disk_usage`, containing one .txt and one .csv file per day. (the single file with a date column is gone).

asnyv:
Resolves #374 

Included some other minor updates like uuid and theming.

This edit is breaking in the way that the old format is not supported, but I don't believe it is an issue.


asnyv:
Moving to #385, as github downtime seems to have messed up the git-tags

PR #380, opened 2020-07-09
Cache testdata

anders-kiaer:


anders-kiaer:
Need to wait on GitHub cache action to be split out in separate save/restore actions.

PR #385, opened 2020-07-13
Update DiskUsage to newer standard file location and format

### LINKED ISSUE #374: DiskUsage no longer compatible with standard

asnyv:
Resolves #374 

Included some other minor updates like uuid and theming.

This edit is breaking in the way that the old format is not supported, but I don't believe it is an issue.

### DIFF
--- a/webviz_subsurface/plugins/_disk_usage.py
+++ b/webviz_subsurface/plugins/_disk_usage.py
@@ -1,5 +1,6 @@
 import pathlib
-from uuid import uuid4
+import datetime
+from typing import Optional
 
 import pandas as pd
 import dash_html_components as html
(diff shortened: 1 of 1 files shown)


webviz_subsurface/plugins/_disk_usage.py
```diff
@@ -101,7 +109,15 @@ def add_webvizstore(self):
 @CACHE.memoize(timeout=CACHE.TIMEOUT)
 @webvizstore
 def get_disk_usage(scratch_dir) -> pd.DataFrame:
-    df = pd.read_csv(scratch_dir / "disk_usage.csv")
-
-    last_date = sorted(list(df["date"].unique()))[-1]
-    return df.loc[df["date"] == last_date]
+    today = datetime.datetime.today()
+    for i in range(0, 7):
```
anders-kiaer, line 113:
```suggestion
    for i in range(7):
```

webviz_subsurface/plugins/_disk_usage.py
```diff
@@ -101,7 +109,15 @@ def add_webvizstore(self):
...
+        date = today - datetime.timedelta(days=i)
+        try:
+            return pd.read_csv(
+                pathlib.Path(scratch_dir)
```
anders-kiaer, line 117: `scratch_dir`is already `Path`?

Issue #326, opened 2020-04-28
Data download / Table view for ReservoirSimulationTimeSeries


HansKallekleiv:
A user wants to see the currently visualized vectors in a table format.
Display tables in webviz (as a separate tab?) or add data download functions.

anders-kiaer:
Also covered by #276.

anders-kiaer:
AFAIK, we haven't seen this user request pop up again (wrt. table) after adding possibility to download data to csc/excel. Closing.

PR #386, opened 2020-07-14
Download data from ReservoirSimulationTimeseries

### LINKED ISSUE #276: Download data from time series

asnyv:
Resolves #276 and #326 (though only download, not table).

Adds functionality to download visualized data from the ReservoirSimulationTimeseries plugin. Data is stored in a zipped folder
as a csvfile.

The code has been somewhat refactored to ease the download functionality + to make the code more modular and reusable.
Some simple performance tests indicate that the new solution is at least not slower than the previous (if anything slightly faster).

Supports:
- [x] Regular ensembles
- [x] Delta ensembles
- [x] Multiple ensembles
- [x] Data per realization
- [x] Ensemble statistics
- [x] Multiple vectors

Some data is intentionally currently not supported for download, as they are not part of the basic dataframe which is exported to csv (they would require custom solutions which I haven't prioritized). Think this PR is useful without them. This includes (at least):
- [ ]  Histogram data (bins) for selected date
- [ ]  Historical vector (if existing)
- [ ]  Observations
- [ ]  Vector metadata

### DIFF
--- a/webviz_subsurface/_utils/simulation_timeseries.py
+++ b/webviz_subsurface/_utils/simulation_timeseries.py
@@ -2,6 +2,7 @@ import warnings
 from typing import Optional
 
 import pandas as pd
+import numpy as np
 
 
 def set_simulation_line_shape_fallback(line_shape_fallback: str) -> str:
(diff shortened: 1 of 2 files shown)


webviz_subsurface/_utils/simulation_timeseries.py
```diff
@@ -118,15 +116,13 @@ def add_fanchart_traces(ens_stat_df, vector, color, legend_group: str, line_shap
...
-            "y": ens_stat_df[(vector, "min")],
+            "y": ens_stat_df[(vector, "mean")],
             "mode": "lines",
-            "fill": "tonexty",
-            "fillcolor": fill_color,
-            "line": {"width": 0, "color": line_color, "shape": line_shape},
+            "line": {"color": line_color, "shape": line_shape},
             "legendgroup": legend_group,
-            "showlegend": False,
+            "showlegend": True,
```
asnyv, line 125: A bit strange, but had to alter the traces to get the right fading of the fanchart. 
On the other hand: the new order fits better with the Plotly reference guide and how fill is expected to behave.

asnyv: Turned out that it is because the previous one used "regular percentiles", but here I have switched P10 and P90 to the oil industry convention (p90 low, p10 high)

webviz_subsurface/_utils/simulation_timeseries.py
```diff
@@ -36,3 +37,104 @@ def get_simulation_line_shape(
     except (AttributeError, KeyError):
         pass
     return line_shape_fallback
+
+
+def calc_series_statistics(df, vectors, refaxis="DATE"):
+    """Calculate statistics for given vectors over the ensembles
+    refaxis is used if another column than DATE should be used to groupby.
+    """
+    # Invert p10 and p90 due to oil industry convention.
```
anders-kiaer, line 46: Since it is "regular" percentiles as of now, maybe we should wait with reverting it? Instead we can change the labels from `P10` and `P90` to `Low (P10)` and `High (P90)` respectively, to improve user communication.

Arguments:
- It is useful to add `Low`/`High` prefixes regardless.
- We don't change current behavior.
- The oil industry is not completely agreeing on a standard. In Equinor, D&W and economical analysis tend to use the common worldwide definition. Since webviz is aiming at being cross disciplinary, maybe the worldwide definition, with clear prefixes, is the best/safest compromise?

PR #389, opened 2020-07-15
Avoid assuming inplace units + support descriptions of any column.

### LINKED ISSUE #388: Avoid assuming units for inplace volumes
Some fields/models might operate with non-metric unit systems.
Dangerous to assume metric in visualizations.

asnyv:
Resolves #388 

This is a proposed solution, not sure if it is the best way to do it, but all units should at least by moved out of the default `volume_terminology.json`. 
The proposed solution also supports descriptions for other column labels than those in the current `volume_terminology.json`,
and any input description will be prioritized if the column label exist in both input and the default group.
Easy to expand from the current` description` and `unit` if additional metadata options are needed in the future. 

A similar solution can e.g. be used for UDQ descriptions in #321, and maybe also to allow metadata for the Eclipse vectors in `ReservoirSimulationTimeseries`, `ReservoirSimulationTimeseriesOneByOne` and `ReservoirSimulationTimeseriesRegional` while using aggregated/csv-exported data (not directly from UNSMRY). 

All of them are basically dependent on whether we think a dictionary stored in json-file is the way to go for such metadata (think we should be fairly consistent once we have decided on something).


asnyv:
blocked by https://github.com/equinor/webviz-config/issues/199

PR #390, opened 2020-07-17
Group b test kommentarer Fra Asgeir

JosteinGj:


webviz_subsurface/plugins/_multiple_regression.py
```diff
@@ -0,0 +1,999 @@
...
+            str(values.min()): {"label": f"{values.min():.2f}"},
+            str(values.max()): {"label": f"{values.max():.2f}"},
+        }
+    )
+
+def theme_layout(theme, specific_layout):
+    layout = {}
+    layout.update(theme["layout"])
+    layout.update(specific_layout)
+    return layout
```
asnyv, line 994: Instead of using this, you can now use the [create_themed_layout](https://github.com/equinor/webviz-config/blob/5f06334c1e98defebb9a5efcb00fde21867822d4/webviz_config/_theme_class.py#L78-L104) method directly on a webviz_theme (then your theme has to be stored as e.g. `self.theme = app.webviz_settings["theme"]` instead of as a `plotly_theme` on the top. This will do a "safe" merge of the specific layout and the existing theme, making sure that nothing is overwritten, which could happen with regular dictionary updates. You can e.g. see how it is used here: https://github.com/equinor/webviz-subsurface/blob/2636ff02c7924bdf50cb80538adb2187de0eee68/webviz_subsurface/plugins/_disk_usage.py#L107

webviz_subsurface/plugins/_multiple_regression.py
```diff
@@ -0,0 +1,999 @@
...
+        )
+    )
+    fig["layout"]["font"].update({"size": 12})
+
+    """Costumizing the hoverer"""
+    fig.update_traces(hovertemplate='%{x}') #x is ticktext
+
+    """Adding arrows to figure"""
+    for i, s in enumerate(sgn):
+        xx = x[i] if num_arrows>1 else x[1]
```
asnyv, line 937: i, s, sgn and xx are not very descriptive (or, maybe i is since it is the output of enumerate). Can make some slightly more descriptive variable names.

webviz_subsurface/plugins/_multiple_regression.py
```diff
@@ -0,0 +1,999 @@
...
+                 f" L {xx-0.025} {s*0.06} " \
+                 f" L {xx-0.07} {s*0.06} " \
+                 f" L {xx} {s*0.08} " \
+                 f" L {xx+0.07} {s*0.06} " \
+                 f" L {xx+0.025} {s*0.06} " \
+                 f" L {xx+0.025} 0 ",
+            line_color="#222A2A",
+            fillcolor="crimson" if p_values[i] < 0.05 else "#606060",
+            line_width=0.6  
+        )
```
asnyv, line 950: Should we have a colorscale on the arrow colors? e.g. a constant scale on p_values from 0 to 1? After all, 0.05 isn't a "magic limit", just a convention. Or is it nicer to keep the same colors as you have on the bars like you do now?

Also: consider for both arrows and bar charts if you can use the theme colors instead of hardcoded ones. The current ones work well in the Equinor theme, but might not look that nice with a different theme.

asnyv:
Also: Run your code with `black _multiple_regression.py` and `pylint _multiple_regression.py`. They will help you with formatting (black tries to modify automatically, pylint gives feedback). These two tests have to be succesful for the automatic checks above to be approved for merge when that time comes :)

PR #394, opened 2020-07-31
Plugins for parameter response multiple regression and parallel coordinates 

### LINKED ISSUE #361: Multiple regression of parameters and responses
### Motivation
The existing parameter-response correlation plugin allows for correlation of only a single parameter with a response ([the ParameterResponseCorrelation plugin](https://github.com/equinor/webviz-subsurface/blob/master/webviz_subsurface/plugins/_parameter_response_correlation.py)). 

It is quite common that just a small number of parameters are the key drivers for a specific response, e.g. the total production over the lifetime of a well. A limited number of key parameters is much easier to handle, both when making a decision, and to sort out which uncertainties that are worth putting more work into. 

### Some suggestions:
- Evaluation / literature search of applicable methods to use. 
    - Consider linear and non-linear methods.
    - Test the most promising ones on the available test data set.
- Implementation and delivery:
    - Implement chosen method(s), either using an existing (well maintained) open source library, or from scratch.
    - User friendly visualization plugin. Should at least feature: selection of response and an "easy to understand" view of key parameter combinations for selected response together with there "score" (quality of fit) 
    - Keep in mind during implementation: How computationally expensive is the chosen method? Can we calculate on request, or do we have to pre-calculate? 
- Additional ideas you might have that can provide additional insight to parameter - response correlations

### LINKED ISSUE #379: Parallel coordinates with response coloring
Parallel coordinate plots are already included in several plugins, but none of them currently allows coloring by a response value. Could either be an optional extension of the current ParameterParallelCoordinates (though responses are currently not a part of this plugin), alternatively a view option in ParameterResponseCorrelation or a standalone plugin like ParameterResponseParallelCoordinates. Standalone might be the cleanest option in the long run? 

It is technically already possible with `TablePlotter` from `webviz-config`, but think this could be so useful that it is worth a standardization. Would also make the possibility clearer for potential users.

VegardOztan:
Resolves #361, resolves #379 

Added plugins for parameter response multiple regression and parallel coordinates.

**_parameter_response_multiple_regression.py** visualizes the results of multiple linear regression of parameters and a chosen response using forward selection to limit the number of terms. 

**_parameter_response_parallel_coordinates.py** genereates parallel coordinates plot colored by response.


JosteinGj:
Further imporovements:

Transforming parameters:

Do it inplace at callback time.
suggested controll panel:
- 2 selectors, one for chosing parameters, the other for choosing the transform to be applied, store and display a list on the form (parameter, transform)

write a function to iterate over selected parameters and transform them. this should be drone before the _gen_model() call in the main callback. Interaction effects are generated first in that call so if transforms are done before they should not affect the algorithm at all.

PR #404, opened 2020-08-13
Implemented a new plugin for visualizing PVT data

### LINKED ISSUE #396: PVT Module : Add new component to WebViz framework
- Import PVT data using opm-module from PyPI
- Use PVT plot in ResInsight for design inspiration 
- Create a PVT plugin for webviz-subsurface

See https://github.com/OPM/ResInsight/issues/6116

rubenthoms:
Also added a new script for extracting PVT data from an Eclipse INIT file. 
A convenience module for reading INIT files was added as 'opm_init_io.py' in '_datainput/'.

This closes #396 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ setup(
             "ReservoirSimulationTimeSeriesRegional = "
             + "webviz_subsurface.plugins:ReservoirSimulationTimeSeriesRegional",
             "RftPlotter =  webviz_subsurface.plugins:RftPlotter",
+            "PvtPlot = webviz_subsurface.plugins:PvtPlot",
         ]
     },
     install_requires=[
@@ -57,6 +58,8 @@ setup(
         "webviz-config>=0.0.55",
         "webviz-subsurface-components>=0.0.23",
         "xtgeo>=2.8",
+        "opm>=2020.10.1",
+        "ecl2df>=0.6.1",
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={"tests": TESTS_REQUIRE},
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ setup(
         ]
     },
     install_requires=[
+        "dash>=1.11",
         "fmu-ensemble>=1.2.3",
         "matplotlib>=3.0",
         "pandas>=0.24",
@@ -58,8 +59,8 @@ setup(
         "webviz-config>=0.0.55",
         "webviz-subsurface-components>=0.0.23",
         "xtgeo>=2.8",
-        "opm>=2020.10.1",
-        "ecl2df>=0.6.1",
+        "opm>=2020.10.1; sys_platform=='linux'",
+        "ecl2df>=0.6.1; sys_platform=='linux'",
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={"tests": TESTS_REQUIRE},
--- a/webviz_subsurface/_datainput/opm_init_io.py
+++ b/webviz_subsurface/_datainput/opm_init_io.py
@@ -1,0 +1,546 @@
+########################################
+#
+#  Copyright (C) 2020-     Equinor ASA
+#
+########################################
+
+# This is giving conflicts with webvizstore
+# from __future__ import annotations
+
+from enum import Enum
+from typing import List
+
+from opm.io.ecl import EclFile
+
+
+class InitFileDefinitions:  # pylint: disable=too-few-public-methods
+    """
+    A namespace for constant definitions for
+    reading Eclipse INIT files.
+    """
+
+    LOGIHEAD_KW = "LOGIHEAD"
+    INTEHEAD_KW = "INTEHEAD"
+    INTEHEAD_UNIT_INDEX = 2
+    INTEHEAD_PHASE_INDEX = 14
+    LOGIHEAD_RS_INDEX = 0
+    LOGIHEAD_RV_INDEX = 1
+    TABDIMS_IBPVTO_OFFSET_ITEM = 6
+    TABDIMS_JBPVTO_OFFSET_ITEM = 7
+    TABDIMS_NRPVTO_ITEM = 8
+    TABDIMS_NPPVTO_ITEM = 9
+    TABDIMS_NTPVTO_ITEM = 10
+    TABDIMS_IBPVTW_OFFSET_ITEM = 11
+    TABDIMS_IBPVTG_OFFSET_ITEM = 13
+    TABDIMS_JBPVTG_OFFSET_ITEM = 14
+    TABDIMS_NRPVTG_ITEM = 15
+    TABDIMS_NPPVTG_ITEM = 16
+    TABDIMS_NTPVTG_ITEM = 17
+    LOGIHEAD_CONSTANT_OILCOMPR_INDEX = 39 - 1
+
+    TABDIMS_IBDENS_OFFSET_ITEM = 18
+    TABDIMS_NTDENS_ITEM = 19
+
+
+class EclPhaseIndex(Enum):
+    Aqua = 0
+    Liquid = 1
+    Vapour = 2
+
+
+class EclPropertyTableRawData:  # pylint: disable=too-few-public-methods
+    """
+    A structure for storing read
+    INIT file data.
+    """
+
+    def __init__(self):
+        self.data = []
+        self.primary_key = []
+        self.num_primary = 0
+        self.num_rows = 0
+        self.num_cols = 0
+        self.num_tables = 0
+
+
+def const_compr_index():
+    return InitFileDefinitions.LOGIHEAD_CONSTANT_OILCOMPR_INDEX
+
+
+def surface_mass_density(ecl_file: EclFile, phase: EclPhaseIndex) -> List[float]:
+    if phase is EclPhaseIndex.Liquid:
+        col = 0
+    elif phase is EclPhaseIndex.Aqua:
+        col = 1
+    elif phase is EclPhaseIndex.Vapour:
+        col = 2
+    else:
+        col = -1
+
+    if col == -1:
+        raise AttributeError("Phase must be Liquid, Aqua or Vapour.")
+
+    tabdims = ecl_file.__getitem__("TABDIMS")
+    tab = ecl_file.__getitem__("TAB")
+
+    start = tabdims[InitFileDefinitions.TABDIMS_IBDENS_OFFSET_ITEM] - 1
+    nreg = tabdims[InitFileDefinitions.TABDIMS_NTDENS_ITEM]
+
+    rho = [tab[start + nreg * (col + 0)], tab[start + nreg * (col + 1)]]
+
+    return rho
+
+
+class VariateAndValues:  # pylint: disable=too-few-public-methods
+    """
+    A structure for holding a variate and
+    multiple covariates.
+    """
+
+    def __init__(self):
+        self.x = 0
+        self.y = []
+
+
+class PvxOBase:
+    """
+    A base class for all fluids.
+    """
+
+    def __init__(self, values: List[VariateAndValues]):
+        self.values = values
+
+    def get_values(self):
+        return self.values
+
+    # pylint: disable=R0201
+    def formation_volume_factor(
+        self, gas_oil_ratio: List[float], pressure: List[float]
+    ) -> List[float]:
+        """
+        Does only return an empty list for now.
+        """
+        if len(gas_oil_ratio) != len(pressure):
+            raise ValueError("rs and po arguments must be of same size")
+        return []
+
+    def viscosity(
+        self, gas_oil_ratio: List[float], pressure: List[float]
+    ) -> List[float]:
+        pass
+
+
+class LiveOil(PvxOBase):
+    pass
+
+
+class WetGas(PvxOBase):
+    pass
+
+
+class DryGas(PvxOBase):
+    pass
+
+
+class WaterImpl(PvxOBase):
+    pass
+
+
+class Implementation:
+    class InvalidArgument(Exception):
+        def __init__(self, message: str):
+            self.message = message
+            super().__init__(message)
+
+    class InvalidType(Exception):
+        def __init__(self):
+            super().__init__("Invalid type. Only live oil/wet gas/water supported.")
+
+    def __init__(self):
+        self.values = []
+
+    def get_table(self, tab_index: int) -> List[VariateAndValues]:
+        if tab_index in range(0, len(self.values)):
+            return self.values[tab_index]
+
+        return None
+
+    def tables(self):
+        return self.values
+
+    @staticmethod
+    def entry_valid(x: float) -> bool:
+        # Equivalent to ECLPiecewiseLinearInterpolant.hpp line 293
+        # or ECLPvtOil.cpp line 458
+        return abs(x) < 1.0e20
+
+
+class Oil(Implementation):
+    def __init__(
+        self, raw: EclPropertyTableRawData, const_compr: bool, rhos: List[float]
+    ):
+        super().__init__()
+        self.rhos = rhos
+        self.values = self.create_pvt_function(raw, const_compr)
+
+    # pylint: disable=unused-argument
+    def create_pvt_function(self, raw: EclPropertyTableRawData, const_compr: bool):
+        if raw.num_primary == 0:
+            raise super().InvalidArgument("Oil PVT table without primary lookup key")
+        if raw.num_cols != 5:
+            raise super().InvalidArgument("PVT table for oil must have five columns")
+        if len(raw.primary_key) != (raw.num_primary * raw.num_tables):
+            raise super().InvalidArgument(
+                "Size mismatch in RS nodes of PVT table for oil"
+            )
+        if len(raw.data) != (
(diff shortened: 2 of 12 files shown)


anders-kiaer:
Thanks @rubenthoms - looks great! Ref. our meeting just now; [here is some documentation](https://github.com/equinor/webviz-config/blob/master/CONTRIBUTING.md#deattaching-data-from-its-original-source) on "portable" `webviz` builds.

setup.py
```diff
@@ -58,6 +59,8 @@
         "webviz-config>=0.0.55",
         "webviz-subsurface-components>=0.0.23",
         "xtgeo>=2.8",
+        "opm>=2020.10.1",
```
anders-kiaer, line 62: Unfortunately, `opm` PyPI binaries I think are not available for iOS and Windows as of now.
```suggestion
        "opm>=2020.10.1; sys_platform=='linux'",
```

webviz_subsurface/_datainput/pvt_data.py
```diff
@@ -0,0 +1,112 @@
...
+
+from webviz_config.common_cache import CACHE
+from webviz_config.webviz_store import webvizstore
+
+from .opm_init_io import Oil, Gas, Water, DryGas
+
+try:
+    from .fmu_input import load_ensemble_set
+except ImportError:
+    from fmu_input import load_ensemble_set
```
anders-kiaer, line 18: Is the need for this `try`/`except` due to `__main__` below?

webviz_subsurface/_datainput/pvt_data.py
```diff
@@ -0,0 +1,112 @@
...
+
+try:
+    import ecl2df
+except ImportError:
+    pass
```
anders-kiaer, line 23: I guess `ecl2df` would fail in the same cases as `.opm_init_io` and `opm.io.ecl` (i.e. missing `opm` binaries)? Add them to same `try`/`except`?

That way we perhaps can offer Mac / Windows users still the csv-option (without failing on import).

webviz_subsurface/_datainput/pvt_data.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+if __name__ == "__main__":
+    read_init_file(
+        ecl2df.EclFiles(
+            "../../webviz-subsurface-testdata/reek_history_match/"
+            "realization-0/iter-0/eclipse/model/5_R001_REEK-0.DATA"
+        )
+    )
+else:
```
anders-kiaer, line 95: I see the value of this, documenting how one of the functions can be used. However, we don't have that for the other submodules in `webviz-subsurface`, instead functions are explained in `docstrings` only (and sometimes through extra unit tests). Maybe follow that same style, and remove this `if` block? Could perhaps also make it possible to remove one of the `try` / `except` blocks above.

webviz_subsurface/_datainput/pvt_data.py
rubenthoms: Yes, I agree that a consistent style makes sense. When removing the `__main__` if case, this `try/except` statement can be removed as well.

setup.py
rubenthoms: Good point. I will add the same requirement for the `ecl2df` package then.

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -0,0 +1,546 @@
+########################################
+#
+#  Copyright (C) 2020-     Equinor ASA
+#
+########################################
+
+# This is giving conflicts with webvizstore
+# from __future__ import annotations
```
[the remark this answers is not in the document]
rubenthoms: I actually wanted to have your opinion on this, that is why I left the comment. In Python 3, it is preferable to use `__future__.annotations` to refer to a class' own type. However, I had to fall back to the Python 2 way of defining the type as a string. This is working of course, but a bit _unpythonian_. I just found a conflict with `webvizstore` which I did not dug deeper into, but it's probably worthwhile to keep this potential conflict in mind. I will remove the comment.

webviz_subsurface/_datainput/pvt_data.py
rubenthoms: Yes, makes totally sense.

webviz_subsurface/_datainput/pvt_data.py
rubenthoms: See comment above. I will remove this `if` case.

webviz_subsurface/_datainput/opm_init_io.py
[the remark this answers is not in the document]
anders-kiaer: Agree, wasn't aware of this conflict 👍 Maybe it is worth a new issue in https://github.com/equinor/webviz-config?

asnyv:
@rubenthoms I tried to run your plugin locally, and I notice that especially for the oil phase (as it has the most complex pvt of the cases), plotting individual realizations generates a lot of traces. For the 4 ensembles*10 realization test case it works somewhat ok (even though the update takes a few seconds), but in a more realistic scenario the number of realizations could easilly be at least an order of magnitude larger, making it hard to use interactively. 

**Therefore:**
_Do you think it is feasible to due some statistics/aggregation to reduce the number of traces?_ 
As you see in the time series plugins, we use fan charts which drastically reduces the number of plotted traces, but that might become messy when each realization generates a large sum of traces like in this pvt module. 

Could be that it is enough to e.g. plot only mean over the realizations in each ensemble? 
We could also have a selector to switch between e.g. mean, p10, p90, max, min. 
Could even have that you could plot single realizations, but then you could only plot 1 realization number (but all ensembles) at the time.

At least we need to find a way to make updates of the view less time consuming than it currently is

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -0,0 +1,543 @@
...
+        tab = ecl_init_file.__getitem__("TAB")
+
+        num_rs = tab_dims[InitFileDefinitions.TABDIMS_NRPVTO_ITEM]
+
+        raw.num_rows = tab_dims[InitFileDefinitions.TABDIMS_NPPVTO_ITEM]
+        raw.num_cols = 5
+        raw.num_tables = tab_dims[InitFileDefinitions.TABDIMS_NTPVTO_ITEM]
+
+        if raw.num_tables == 0:
+            return None
```
asnyv, line 270: Same thing
```suggestion
    def from_ecl_init_file(ecl_init_file: EclFile) -> Optional["Oil"]:
        logihead = ecl_init_file.__getitem__(InitFileDefinitions.LOGIHEAD_KW)
        is_const_compr = bool(logihead[const_compr_index()])

        raw = EclPropertyTableRawData()

        tab_dims = ecl_init_file.__getitem__("TABDIMS")
        tab = ecl_init_file.__getitem__("TAB")

        num_rs = tab_dims[InitFileDefinitions.TABDIMS_NRPVTO_ITEM]

        raw.num_rows = tab_dims[InitFileDefinitions.TABDIMS_NPPVTO_ITEM]
        raw.num_cols = 5
        raw.num_tables = tab_dims[InitFileDefinitions.TABDIMS_NTPVTO_ITEM]

        if raw.num_tables == 0:
            return None
```

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -0,0 +1,543 @@
...
+        tab = ecl_init_file.__getitem__("TAB")
+
+        num_rv = tab_dims[InitFileDefinitions.TABDIMS_NRPVTG_ITEM]
+        num_pg = tab_dims[InitFileDefinitions.TABDIMS_NPPVTG_ITEM]
+
+        raw.num_cols = 5
+        raw.num_tables = tab_dims[InitFileDefinitions.TABDIMS_NTPVTG_ITEM]
+
+        if raw.num_tables == 0:
+            return None
```
asnyv, line 437: You guessed it ;) 
```suggestion
    def from_ecl_init_file(ecl_init_file: EclFile) -> Optional["Gas"]:
        intehead = ecl_init_file.__getitem__(InitFileDefinitions.INTEHEAD_KW)
        intehead_phase = intehead[InitFileDefinitions.INTEHEAD_PHASE_INDEX]

        if (intehead_phase & (1 << 2)) == 0:
            return None

        raw = EclPropertyTableRawData()

        tab_dims = ecl_init_file.__getitem__("TABDIMS")
        tab = ecl_init_file.__getitem__("TAB")

        num_rv = tab_dims[InitFileDefinitions.TABDIMS_NRPVTG_ITEM]
        num_pg = tab_dims[InitFileDefinitions.TABDIMS_NPPVTG_ITEM]

        raw.num_cols = 5
        raw.num_tables = tab_dims[InitFileDefinitions.TABDIMS_NTPVTG_ITEM]

        if raw.num_tables == 0:
            return None
```

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -0,0 +1,543 @@
...
+
+        return ret
+
+    @staticmethod
+    def from_ecl_init_file(ecl_init_file: EclFile) -> "Water":
+        intehead = ecl_init_file.__getitem__(InitFileDefinitions.INTEHEAD_KW)
+        intehead_phase = intehead[InitFileDefinitions.INTEHEAD_PHASE_INDEX]
+
+        if (intehead_phase & (1 << 2)) == 0:
+            return None
```
asnyv, line 520:
```suggestion
    def from_ecl_init_file(ecl_init_file: EclFile) -> Optional["Water"]:
        intehead = ecl_init_file.__getitem__(InitFileDefinitions.INTEHEAD_KW)
        intehead_phase = intehead[InitFileDefinitions.INTEHEAD_PHASE_INDEX]

        if (intehead_phase & (1 << 2)) == 0:
            return None
```

webviz_subsurface/plugins/_pvt_plot.py
```diff
@@ -0,0 +1,640 @@
...
+            (
+                load_pvt_dataframe,
+                [
+                    {
+                        "ensemble_paths": self.ensemble_paths,
+                        "ensemble_set_name": "EnsembleSet",
+                    }
+                ],
+            )
+        ]
```
asnyv, line 378: This only supports the `ecl2df` scenario. Need to add some logic to generalize it. The suggestion assumes that you add something like `self.read_from_init_file = read_from_init_file` in `PvtPlot`'s `init`
```suggestion
    def add_webvizstore(self):
        return (
            [
                (
                    load_pvt_dataframe,
                    [
                        {
                            "ensemble_paths": self.ensemble_paths,
                            "use_init_file": self.read_from_init_file,
                        }
                    ],
                )
            ]
            if self.pvt_relative_file_path is None
            else [
                (
                    load_csv,
                    [
                        {
                            "ensemble_paths": self.ensemble_paths,
                            "csv_file": self.pvt_relative_file_path,
                        }
                    ],
                )
            ]
        )
```

webviz_subsurface/plugins/_pvt_plot.py
```diff
@@ -0,0 +1,640 @@
...
+                ensemble
+            ]
+            for ensemble in ensembles
+        }
+
+        self.plotly_theme = app.webviz_settings["theme"].plotly_theme
+
+        self.pvt_relative_file_path = pvt_relative_file_path
+
+        if self.pvt_relative_file_path is not None:
```
asnyv, line 83: could be nicer to switch the order of the if-statement to get `if self.pvt_relative_file_path is None` instead, but that is just a preference :wink:

webviz_subsurface/plugins/_pvt_plot.py
```diff
@@ -0,0 +1,640 @@
...
+    @property
+    def ensembles(self) -> List[str]:
+        return list(self.pvt_data_frame["ENSEMBLE"].unique())
+
+    @property
+    def pvtnums(self) -> List[str]:
+        return list(self.pvt_data_frame["PVTNUM"].unique())
+
+    @property
+    def ensemble_colors(self) -> List[str]:
```
asnyv, line 141: I think it is something like this?
```suggestion
    def ensemble_colors(self) -> Dict[str, List[str]]:
```

webviz_subsurface/plugins/_pvt_plot.py
```diff
@@ -0,0 +1,640 @@
...
+        return {
+            ensemble: self.plotly_theme["layout"]["colorway"][
+                self.ensembles.index(ensemble)
+            ]
+            for ensemble in self.ensembles
+        }
+
+    @property
+    def pvtnum_colors(self) -> List[str]:
```
asnyv, line 150: same as above

webviz_subsurface/plugins/_pvt_plot.py
```diff
@@ -0,0 +1,640 @@
...
+            Output(self.uuid("graph"), "figure"),
+            [
+                Input(self.uuid("color_by"), "value"),
+                Input(self.uuid("phase"), "value"),
+                Input(self.uuid("ensemble"), "value"),
+                Input(self.uuid("pvtnum"), "value"),
+            ],
+        )
+        def _update_graph(
+            color_by: str, phase: str, ensembles: List[str], pvtnums: List[str]
```
asnyv, line 297: As you support conversion from a str to a list further down
```suggestion
            color_by: str, phase: str, ensembles: Union[List[str], str], pvtnums: Union[List[str], str]
```

webviz_subsurface/plugins/_pvt_plot.py
```diff
@@ -0,0 +1,640 @@
...
+            )
+            self.pvt_data_frame = self.pvt_data_frame.rename(
+                str.upper, axis="columns"
+            ).rename(columns={"TYPE": "KEYWORD", "RS": "GOR"})
+
+        self.set_callbacks(app)
+
+    @property
+    def phases(self) -> List[str]:
+        return PvtPlot.PHASES.keys()
```
asnyv, line 130: Don't think this is a List[str] but rather of `dict_keys` class?

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -0,0 +1,543 @@
...
+            super().__init__("Invalid type. Only live oil/wet gas/water supported.")
+
+    def __init__(self):
+        self.values = []
+
+    def get_table(self, tab_index: int) -> List[VariateAndValues]:
+        if tab_index in range(0, len(self.values)):
+            return self.values[tab_index]
+
+        return None
```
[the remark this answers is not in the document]
rubenthoms: Absolutely correct ;)

webviz_subsurface/plugins/_pvt_plot.py
rubenthoms: Good catch, that is definitely an important fix.

webviz_subsurface/plugins/_pvt_plot.py
rubenthoms: Agree, it somehow reads nicer the other way around ;)

webviz_subsurface/plugins/_pvt_plot.py
rubenthoms: True.

webviz_subsurface/plugins/_pvt_plot.py
rubenthoms: Correct, I somehow was still in the Python 2.x world in my head ;) I changed it to `list(PvtPlot.PHASES.keys())` now.

asnyv:
Did you btw see the performance comment before the regular review @rubenthoms ?

rubenthoms:
@asnyv Yes, I have seen it. Thanks for the very valuable suggestions :-) We just have to discuss how much time we are still having available for this project. I think Magne Sjaastad wanted to write to Anders to discuss this.

magnesj:
@asnyv , when do you think you have time to complete the review as @rubenthoms requested? He has fixed all code-related issues based on your first review, but we have not done anything on performance.

asnyv:
> @asnyv , when do you think you have time to complete the review as @rubenthoms requested? He has fixed all code-related issues based on your first review, but we have not done anything on performance.

@magnesj I can look at it right away. As the comments I had were quite straight forward, I would guess it is pretty much ok ;) You had a chat regarding the performance as far as I understood?

magnesj:
@anders-kiaer and I had a short discussion and agreed to implement the following:

- plot PVT data for the first model in the collection of models
- if data is read from csv-files, detect if there are differences in PVT-data between models
- if differences are detected, write a warning to console : "Detected differences in PVT-data. Plotting of PVT uncertainty is not supported"
- if time, also display this warning in frontend

We plan to update this PR with these changes this week.

rubenthoms:
@asnyv Thanks for the detailed review comments. I implemented a new version which collects PVT data from all realizations for one ensemble, removes all duplicates and then only uses ensembles that are not equal to others that have already been added before. I experienced a significant speed-up with the reek test data, hope this gives a better experience also with larger amounts of data.

asnyv:
Seems like something strange happened with the latest commit @rubenthoms? Might be related to your rebase to master. Additional files have been included.

rubenthoms:
@asnyv Yes, I did not fully understand what was going on. I'm closing this pull request in favour of #411.

PR #408, opened 2020-08-19
Improved verification of observation file for time series observations.

### LINKED ISSUE #392: Give user friendly error message when yaml obs file is invalid

asnyv:
Closes #392

### DIFF
--- a/webviz_subsurface/_utils/simulation_timeseries.py
+++ b/webviz_subsurface/_utils/simulation_timeseries.py
@@ -1,8 +1,12 @@
+import sys
 import warnings
 from typing import Optional
+from pathlib import Path
 
 import pandas as pd
 import numpy as np
+import yaml
+from webviz_config.utils import terminal_colors
 
 
 def set_simulation_line_shape_fallback(line_shape_fallback: str) -> str:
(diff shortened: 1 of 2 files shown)


webviz_subsurface/_utils/simulation_timeseries.py
```diff
@@ -151,3 +153,72 @@ def hex_to_rgb(hex_string, opacity=1):
     rgb = [int(hex_string[i : i + hlen // 3], 16) for i in range(0, hlen, hlen // 3)]
     rgb.append(opacity)
     return f"rgba{tuple(rgb)}"
+
+
+# pylint: disable=too-many-branches
+def check_and_format_observations(obsfile: Path):
+    with open(obsfile, "r") as stream:
+        obsfile_data = yaml.safe_load(stream)
```
anders-kiaer, line 161: The user that reported this error actually failed on this line. So maybe include that as well in a `try`? (He had wrong indentation or something making it an invalid yaml file).

asnyv: @anders-kiaer Hehe, was considering to wrap it, but then I thought the error message I got out wasn't that bad. 
But ok, can do something similar to what we have in webviz config for the yaml-input. 

Do we want to color these as well? Similar to build?

anders-kiaer: You choose. I don't think it is necessary to color it. He though just something was wrong/buggy due to the long traceback, and difficult (from user point of view) to see from the traceback that it was the obs.yaml file that was invalid (the traceback only shows the variable names etc.)

anders-kiaer (approved):

PR #415, opened 2020-08-24
Use built in xml parsing

HansKallekleiv:


anders-kiaer (approved):
Squash before merge, otherwise LGTM! 🍰

PR #420, opened 2020-08-27
Use new Leaflet Map component

### LINKED ISSUE #397: Use new Map component
The following plugins should replace use of `webviz_subsurface_components.LayeredMap` with `webviz_subsurface_components.LeafletMap`:

- SurfaceViewer
- SurfaceWithGridCrossSection
- SurfaceWithSeismicCrossSection
- WellCrossSection
- WellCrossSectionFMU

HansKallekleiv:
Closes #397


anders-kiaer:
Nice @HansKallekleiv! 😎🚀 

Should we just 🔪 away `webviz_subsurface.image_processing.get_colormap` now? I guess that is not used anymore by `webviz-subsurface`? 

Potentialy we can now probably also get rid of some MB of uncompressed `matplotlib` install files as well in the 🐳 Docker image now? I think our actual usage of `matplotlib` is small, and especially after this PR (as far as I can see only [`ListedColorMap`](https://matplotlib.org/api/_as_gen/matplotlib.colors.ListedColormap.html), which I guess we easily can do ourselves or using `plotly` 🐍  functionality?). I'm open for removing `maptlotlib` in a separate PR (doesn't have to be this one).

anders-kiaer (approved):
Nice. Squash some commits, and then this is ready to go! 🚀 👍

PR #428, opened 2020-09-08
Persist user settings in plugins across sessions

HansKallekleiv:
### Contributor checklist

   - [x] Use persistence in dash_core_components and webviz_core_components.Select
   - [x] Store `storage_type=session` in `dcc.Store` components to keep data when as long as the browser window is open.


anders-kiaer (approved):
Nice work 🚀🍾

PR #432, opened 2020-09-14
Only strip on match of full string in unit_vec

### LINKED ISSUE #431: Vector starting with A,V,G,I,N,T,V or L are stripped in ReservoirSimulationTimeSeries
**Describe the bug**
Vector starting with A,V,G,I,N,T,V or L are stripped in ReservoirSimulationTimeSeries and crashes due to missing column in DataFrame

**How to reproduce**
1. Go to https://webviz-subsurface-example.azurewebsites.net/
2. Click on Simulation time series
3. Choose vector GGPR:OP

**Additional context**
`lstrip`'ed [here](https://github.com/equinor/webviz-subsurface/blob/65a1075f5c054214c8bbfcae27dc256a0af4831f/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py#L618) which strippes letter by letter and not the full string. Should be switched to match the whole string.

asnyv:

---

### Contributor checklist

- [X] :tada: This PR closes #431 .

### DIFF
--- a/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
+++ b/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
@@ -1,4 +1,5 @@
 # pylint: disable=too-many-lines
+import sys
 from pathlib import Path
 import json
 
@@ -615,7 +616,14 @@ folder, to avoid risk of not extracting the right data.
             # Titles for subplots
             titles = []
             for vec in vectors:
-                unit_vec = vec.lstrip("AVG_").lstrip("INTVL_")
+                if sys.version_info >= (3, 9):
+                    unit_vec = vec.removeprefix("AVG_").removeprefix("INTVL_")
+                else:
+                    unit_vec = (
+                        vec[4:]
+                        if vec.startswith("AVG_")
+                        else (vec[6:] if vec.startswith("INTVL_") else vec)
+                    )
                 if self.smry_meta is None:
                     titles.append(simulation_vector_description(vec))
                 else:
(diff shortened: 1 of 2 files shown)


webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -615,7 +615,11 @@ def _update_graph(
             # Titles for subplots
             titles = []
             for vec in vectors:
-                unit_vec = vec.lstrip("AVG_").lstrip("INTVL_")
+                unit_vec = (
+                    vec[4:]
+                    if vec.startswith("AVG_")
+                    else (vec[6:] if vec.startswith("INTVL_") else vec)
+                )
```
anders-kiaer, line 622: Maybe we can change to this in order to both increase readability, but also to remember that this can be simplified when (in some years) Python 3.8 support is dropped.
```suggestion
                if sys.version_info >= (3.9):
                    unit_vec = vec.removeprefix("AVG_").removeprefix("INTVL_")
                else:
                    unit_vec = (
                        vec[4:]
                        if vec.startswith("AVG_")
                        else (vec[6:] if vec.startswith("INTVL_") else vec)
                    )
```

anders-kiaer (changes requested):
LGTM! One small comment. Update `CHANGELOG.md` as well?

asnyv:
> LGTM! One small comment. Update `CHANGELOG.md` as well?

Added a short description in the changelog

anders-kiaer (approved):
Nice fix 🚀🎉

PR #434, opened 2020-09-16
Improve map hillshading and color scale

### LINKED ISSUE #429: SurfaceViewerFMU : Surfaces are not truncated to specified min/max
Only the color legend is updated.

HansKallekleiv:
- [x] :tada: This PR closes #429 .
   - [x] Adjust map colors to user-specified min/max values
   - [x] Better default hillshading settings
   - [x] Update CHANGELOG


anders-kiaer (changes requested):
Add a changelog entry, as well as changing back to `equinor` and `master` in GitHub workflow, other than that LGTM! 🐛👍

PR #440, opened 2020-09-18
'Remove' opm and ecl2df dependencies in setup.py

anders-kiaer:
We unfortunately can't include `libecl` and `opm` as formal dependencies yet due to limitations in internal distribution system :disappointed:


HansKallekleiv (commented):
Just a question, should we provide a provide a warning or document that these packages are required?

anders-kiaer:
> Just a question, should we provide a provide a warning or document that these packages are required?

Good idea - added a temporary warning in `README.md` until they can be added formally to `setup.py` again.

HansKallekleiv (approved):
lgtm 👍

PR #466, opened 2020-10-08
Maintenance setup.py + __init__.py

anders-kiaer:
- Make list alphabetic.
- Don't split long lines with `+` etc. (less readable).


asnyv (approved):
👍🎉

PR #478, opened 2020-10-28
AHM analysis plugin

mareco701:
This is to add a new plugin to visualize assisted history matching analysis (how much update in parameters observations trigger, and which observations trigger an update of a parameter when all observations are considered)
---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


anders-kiaer:
Thanks for the PR @mareco701. Changed the status of this to "draft" while waiting for things to go :green_circle:.

Looks like a module is currently missing according to the logs?
`ModuleNotFoundError: No module named 'data_analysis_test1'`

setup.py
```diff
@@ -53,6 +53,7 @@
             "SurfaceWithSeismicCrossSection = webviz_subsurface.plugins:SurfaceWithSeismicCrossSection",
             "WellCrossSection = webviz_subsurface.plugins:WellCrossSection",
             "WellCrossSectionFMU = webviz_subsurface.plugins:WellCrossSectionFMU",
+            "AssistedHistoryMatchingAnalysis = data_analysis_test1.plugins:AssistedHistoryMatchingAnalysis",
```
anders-kiaer, line 56:
```suggestion
            "AssistedHistoryMatchingAnalysis = webviz_subsurface.plugins:AssistedHistoryMatchingAnalysis",
```
...and you should then also add it to `webviz_subsurface/plugins/__init__.py` 🙂

webviz_subsurface/plugins/_ahm_analysis_plugin.py
```diff
@@ -0,0 +1,435 @@
...
+from webviz_config import WebvizPluginABC
+
+
+import plotly.graph_objs as go
+
+import plotly.express as px
+import dash_table
+
+
+class AHMAnalysisPlugin(WebvizPluginABC):
```
anders-kiaer, line 19: The plugin name here should be consistent with what is used in `setup.py`.

anders-kiaer:
Looks good :+1:  Changing temporarily these lines https://github.com/equinor/webviz-subsurface/blob/5f5dc6596ce2a4122b9c53c1698a1170b1f47d5d/.github/workflows/subsurface.yml#L83-L86 to `mareco701` and `inc_test_for_ahmanalysis` would help in the last part of the review process by automatically testing against your new testdata PR (then we simply change those two lines back before `merge`).

anders-kiaer:
This looks ready for review now @mareco701 :+1: :tada:

webviz_subsurface/plugins/_ahm_analysis_plugin.py
```diff
@@ -0,0 +1,434 @@
...
+                    param = paraml[0]
+            fig = go.Figure()
+            fig.add_trace(go.Histogram(x=prior_df[param], name="prior", nbinsx=10))
+            fig.add_trace(go.Histogram(x=post_df[param], name="update", nbinsx=10))
+            fig.update_layout(
+                title="Parameter distribution for observation "
+                + obs
+                + " ("
+                + str(active_info.at["ratio", obs])
+                + ")",
```
anders-kiaer, line 311:
```suggestion
                title=f"Parameter distribution for observation {obs} ({active_info.at["ratio", obs]})",
```

webviz_subsurface/plugins/_ahm_analysis_plugin.py
```diff
@@ -0,0 +1,434 @@
...
+        @app.callback(
+            Output(
+                component_id=self.ids("generate_table"), component_property="children"
+            ),
+            [
+                Input(
+                    component_id=self.ids("inputpath_id"), component_property="value"
+                ),
+                Input(component_id=self.ids("choice_id"), component_property="value"),
+            ],
```
anders-kiaer, line 327:
```suggestion
            Output(self.ids("generate_table"), component_property="children"),
            Input(self.ids("inputpath_id"), component_property="value"),
            Input(self.ids("choice_id"), component_property="value"),
```

webviz_subsurface/plugins/_ahm_analysis_plugin.py
```diff
@@ -0,0 +1,434 @@
...
+            if active_obs_info[0] == "0":
+                zz_data[yid][xid] = None
+    return zz_data
+
+
+def set_inputfilter(input_filter):
+    """set the input filter to show all data if empty"""
+    if input_filter == "":
+        input_filter = "_"
+    return input_filter
```
anders-kiaer, line 434:
```suggestion
    return "_" if input_filter == "" else input_filter
```

anders-kiaer (changes requested):
Nice plugin @mareco701. 

> Give path to input data: 

Should we remove this input box in the frontend, since it is already defined in the `yaml` file? In addition asking for file input will not work very well in :cloud: setting.

> Show default parameters distribution
> Show transformed parameters distribution (if available)

Then maybe at plugin startup you already know if the transformed distribution is available, and can hide (or disable/grey out) this option in order to simplify the UI? Since it is binary choice (default or transformed) maybe it can be a check box instead of two radio buttons?

Add a guided tour explaining what the uses sees? See https://github.com/equinor/webviz-subsurface/blob/f9062a639974518c4e51c726b57a5c0111810bb9/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py#L256 for an example from another plugin.

> Filter observations: 
> Filter parameters: 

Maybe they should use e.g. `wcc.Select` instead of text input box? Both to make it easier to select among other options, and to ensure only valid options are entered (currently writing an non-existing parameter in the parameter filter results in an `IndexError`). See https://webviz-subsurface-example.azurewebsites.net/parameter-parallel-coordinates for an example using `wcc.Select`.

webviz_subsurface/plugins/_ahm_analysis_plugin.py
mareco701: not changed since it made black fail

HansKallekleiv (approved):
Looks good! :+1:

anders-kiaer (approved):

PR #480, opened 2020-10-29
Add OK filtering to scratch_ensemble (applies to InplaceVolumes and S…

### LINKED ISSUE #475: scratch_ensemble used in SubsurfaceMap not filtered on OK
[scratch_ensemble](https://github.com/equinor/webviz-subsurface/blob/77dfbc6f151a8d705e7af28280486be684f5d3d3/webviz_subsurface/_datainput/fmu_input.py#L11-L13) (CACHE wrapped fmu-ensemble ScratchEnsemble) is not filtering on OK file, crashes when simulations have crashed. Either filter this one as well, or wrap the ensemble into an ensembleset, and use the existing load_ensemble_set (and remove scratch_ensemble). Might be slighly less maintenance?

Same function is implemented in [_data_input/inplace_volumes](https://github.com/equinor/webviz-subsurface/blob/77dfbc6f151a8d705e7af28280486be684f5d3d3/webviz_subsurface/_datainput/inplace_volumes.py#L9-L11), but here errors are silently ignored unless all files are missing. Therefore not creating issues. Think we should consider to rather use `load_ensemble_set` and then iterate over the ensembles instead though.

A last thing is if we should let the filtering option available in [load_ensemble_set](https://github.com/equinor/webviz-subsurface/blob/77dfbc6f151a8d705e7af28280486be684f5d3d3/webviz_subsurface/_datainput/fmu_input.py#L16-L30) be avialable for all the methods that call `load_ensemble_set` in [fmu_input](https://github.com/equinor/webviz-subsurface/blob/master/webviz_subsurface/_datainput/fmu_input.py)?  

Currently some do, like [load_parameters](https://github.com/equinor/webviz-subsurface/blob/77dfbc6f151a8d705e7af28280486be684f5d3d3/webviz_subsurface/_datainput/fmu_input.py#L33-L39), but could make sense to add it for all methods? Will maybe be a breaking change for most existing portables due to new input in webvizstore functions...

asnyv:
### Contributor checklist

- [X] :tada: This PR closes #475 
   - [X] Add filter to scratch_ensemble method in _datainput.fmu_input
   - [X] Make _datainput.inplace_volumes dependent on fmu_input.scratch_ensemble
   - [X] Use scratch_ensemble with filter in load_ensemble_set to make it slightly more compact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#450](https://github.com/equinor/webviz-subsurface/pull/450) - Flipped colormap for subsurface maps (such that deeper areas get darker colors). Also fixed hill shading such that input values are treated as depth, not positive elevation.
 - [#459](https://github.com/equinor/webviz-subsurface/pull/459) - Bug fix in ReservoirSimulationTimeSeries. All `History` traces are now toggled when clicking `History` in the legend.
 - [#474](https://github.com/equinor/webviz-subsurface/pull/474) - Bug fix in ParameterCorrelation. Constant parameters are now removed if `drop_constants` is set to `True`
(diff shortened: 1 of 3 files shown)


HansKallekleiv (approved):
LGTM 👍

PR #481, opened 2020-10-29
RFT-plotter: Add support for MD, and make ECLIPSE RFT data optional.

tnatt:
Support using the RFT-plotter with only ERT data, by removing ECPLIPSE RFT data as a requirement.

Added support for plotting MD values on "Depth vs pressure plot". In most cases the simulated MD will be plotted only for the ERT observation points. In order to plot simulated MD values as lines from the ECPLIPSE RFT data a "CONMD" column needs to be present in the rft.csv file, and only wells that have a unique MD value per RFT datapoint will be plotted (else ERT points are shown).

Add support for new standard format for rft_ert.csv which will have column name "is_active" instead of  "isactive". 

---

### Contributor checklist

   - [x] Support MD
   - [x] Make ECLIPSE RFT data optional


HansKallekleiv (approved):
LGTM 👍 
Remember to change back testdata repo

webviz_subsurface/plugins/_rft_plotter/_formation_figure.py
```diff
@@ -128,47 +211,63 @@ def add_ert_observed(self, date):
...
+                                "name": ensemble,
+                                "showlegend": i == 0,
+                                "legendgroup": ensemble,
+                            }
+                        )
+            if linetype == "fanchart":
+                for ensemble, ensdf in self.simdf.groupby("ENSEMBLE"):
+                    dframe = interpolate_depth(ensdf)
+                    quantiles = [10, 90]
+                    dframe = dframe.drop(columns=["REAL"]).groupby("DEPTH")
```
anders-kiaer, line 253:
```suggestion
                    quantiles = [10, 90]
                    dframe = interpolate_depth(ensdf).drop(columns=["REAL"]).groupby("DEPTH")
```

webviz_subsurface/plugins/_rft_plotter/rft_plotter.py
```diff
@@ -90,12 +91,10 @@ class RftPlotter(WebvizPluginABC):
...
 
-?> Only TVD values are supported, plan to support MD values in a later release.
-
-The `rft_ert.csv` file currently lacks a standardized method of generation. A \
-**temporary** script can be found [here]\
-(https://github.com/equinor/webviz-subsurface-testdata/\
-blob/b8b7f1fdd3abc505b137b587dcd9e44bbcf411c9/preprocessing_scripts/ert_rft.py).
+The `rft_ert.csv` file can be generated by running the "MERGE_RFT_ERTOBS" forward model in ERT, \
+this will collect ERT RFT observations and merge with CSV output from the "GENDATA_RFT" forward \
+model. [ERT docs] (https://fmu-docs.equinor.com/docs/ert/reference/\
```
anders-kiaer, line 96:
```suggestion
model. [ERT docs](https://fmu-docs.equinor.com/docs/ert/reference/\
```

webviz_subsurface/plugins/_rft_plotter/rft_plotter.py
```diff
@@ -219,22 +212,34 @@ def add_webvizstore(self):
...
+                        load_csv,
+                        [
+                            {
+                                "ensemble_paths": self.ens_paths,
+                                "csv_file": "share/results/tables/rft.csv",
+                            },
+                        ],
+                    )
+                )
+            except KeyError:
```
anders-kiaer, line 240: What would be the source code raising the `KeyError` when that happens?

webviz_subsurface/plugins/_rft_plotter/rft_plotter.py
```diff
@@ -774,30 +809,82 @@ def _update_map(ensemble, sizeby, colorby, dates):
...
+                Output(self.uuid("linetype"), "value"),
+            ],
+            [
+                Input(self.uuid("depth_option"), "value"),
+            ],
+            [
+                State(self.uuid("linetype"), "value"),
+                State(self.uuid("well"), "value"),
+                State(self.uuid("date"), "value"),
+            ],
```
anders-kiaer, line 854:
```suggestion
            Output(self.uuid("linetype"), "options"),
            Output(self.uuid("linetype"), "value"),
            Input(self.uuid("depth_option"), "value"),
            State(self.uuid("linetype"), "value"),
            State(self.uuid("well"), "value"),
            State(self.uuid("date"), "value"),
```

anders-kiaer (changes requested):
:+1: Looks good to me. Some minor comments/suggestions only. `testdata` PR is also merged now so can change back GitHub action workflow.

Some strings are using `tvd` and `md`, while others are using upper case. Would it make sense to choose one form, e.g. upper case, or are there arguments for keeping a mixture?

.github/workflows/subsurface.yml
```diff
@@ -80,10 +80,10 @@ jobs:
       env:
         # If you want the CI to (temporarily) run against your fork of the testdada,
         # change the value her from "equinor" to your username.
-        TESTDATA_REPO_OWNER: equinor
+        TESTDATA_REPO_OWNER: tnatt
```
anders-kiaer, line 83:
```suggestion
        TESTDATA_REPO_OWNER: equinor
```

webviz_subsurface/plugins/_rft_plotter/rft_plotter.py
tnatt: If the rft.csv file is not found in any realization fmu.ensemble will raise a KeyError. If this happens we should not append that function to Webviz Store.

anders-kiaer (approved):

PR #482, opened 2020-10-29
Drop NaN in delta profiles instead of replacing with zeros

### LINKED ISSUE #472: Deltas when realization is missing in one ensemble is set to 0
**Describe the bug**
Deltas in ReservoirSimulationTimeSeries where one realization is missing is set to 0, [see end of this function](https://github.com/equinor/webviz-subsurface/blob/77dfbc6f151a8d705e7af28280486be684f5d3d3/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py#L1012-L1026). This affects statistics.

**Expected behavior**
Should filter out realizations which don't exist in both ensembles.

asnyv:
Note that in the case of different ensemble sizes, only matching realization indices will be kept (no other logical way to handle it as far as I know).
- [X] :tada: This PR closes #472 
   - [X] Switch fillna(0) with dropna(axis=0, how="any") (these are also default options, but in case default behavior changes...)
   - [X] Handle case of empty delta (just skipping, no message, error or warning, but user should know this). Result is empty plot.


anders-kiaer (approved):
:+1:

PR #486, opened 2020-11-08
Use realization number instead of dataframe index for tagging realizations

HansKallekleiv:


tnatt (approved):
LGTM 👍

anders-kiaer (changes requested):
LGTM 👍 Add a `Fix` changelog entry before merge?

anders-kiaer (approved):

PR #509, opened 2020-11-30
Improvements to parameter analysis

tnatt:
Several improvements to parameter analysis plugin

---

### Contributor checklist

   - [x] Added hoverinfo and improvements to layout on parameter distributions plots
   - [x] Refactored color_figure
   - [x] Optimized callback logic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - [#529](https://github.com/equinor/webviz-subsurface/pull/529) - Added support for PVDO and PVTG to PVT plot and to respective data modules.
+- [#509](https://github.com/equinor/webviz-subsurface/pull/509) - Added descriptive hoverinfo to  `ParameterAnalysis`. Average and standard deviation of parameter value
+for each ensemble shown on mouse hover over figure. Included dynamic sizing of plot titles and plot spacing to optimize the appearance of plots when many parameters are plotted.
 
 ## [0.1.6] - 2020-11-30
 ### Fixed
--- a/webviz_subsurface/plugins/_parameter_analysis/controllers/parameter_response_controller.py
+++ b/webviz_subsurface/plugins/_parameter_analysis/controllers/parameter_response_controller.py
@@ -68,14 +68,12 @@ def parameter_response_controller(parent, app):
             raise PreventUpdate
         ctx = dash.callback_context.triggered[0]["prop_id"].split(".")[0]
 
+        initial_run = timeseries_fig is None
         color = options["color"] if options["color"] is not None else "#007079"
         daterange = parent.vmodel.daterange_for_plot(vector=vector)
 
         # Make timeseries graph
-        if (
-            relevant_ctx_for_plot(parent, ctx, plot="timeseries_fig")
-            or timeseries_fig is None
-        ):
+        if relevant_ctx(parent, ctx, operation="timeseries_fig") or initial_run:
             timeseries_fig = update_timeseries_graph(
                 parent.vmodel,
                 ensemble,
@@ -84,20 +82,19 @@ def parameter_response_controller(parent, app):
                 real_filter=None,
             )
 
-        vectors_filtered = filter_vectors(
-            parent, vector_type_filter, vector_item_filters
-        )
-        if vector not in vectors_filtered:
-            vectors_filtered.append(vector)
-
-        merged_df = merge_parameter_and_vector_df(
-            parent, ensemble, vectors_filtered, date
-        )
+        if parent.uuid("plot-options") not in ctx or initial_run:
+            vectors_filtered = filter_vectors(
+                parent, vector_type_filter, vector_item_filters
+            )
+            if vector not in vectors_filtered:
+                vectors_filtered.append(vector)
+            merged_df = merge_parameter_and_vector_df(
+                parent, ensemble, vectors_filtered, date
+            )
 
         # Make correlation figure for vector
         if options["autocompute_corr"] and (
-            relevant_ctx_for_plot(parent, ctx, plot="vector_correlation")
-            or corr_v_fig is None
+            relevant_ctx(parent, ctx, operation="vector_correlation") or initial_run
         ):
             corr_v_fig = make_correlation_figure(
                 merged_df, response=vector, corrwith=parent.pmodel.parameters
@@ -111,8 +108,7 @@ def parameter_response_controller(parent, app):
 
         # Make correlation figure for parameter
         if options["autocompute_corr"] and (
-            relevant_ctx_for_plot(parent, ctx, plot="parameter_correlation")
-            or corr_p_fig is None
+            relevant_ctx(parent, ctx, operation="parameter_correlation") or initial_run
         ):
             corr_p_fig = make_correlation_figure(
                 merged_df, response=parameter, corrwith=vectors_filtered
@@ -121,7 +117,7 @@ def parameter_response_controller(parent, app):
         corr_p_fig = color_corr_bars(corr_p_fig, vector, color, options["opacity"])
 
         # Create scatter plot of vector vs parameter
-        if relevant_ctx_for_plot(parent, ctx, plot="scatter") or scatter_fig is None:
+        if relevant_ctx(parent, ctx, operation="scatter") or initial_run:
             scatter_fig = update_scatter_graph(merged_df, vector, parameter, color)
 
         scatter_fig = scatter_fig_color_update(scatter_fig, color, options["opacity"])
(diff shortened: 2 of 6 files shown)


webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
```diff
@@ -206,39 +206,94 @@ def make_grouped_plot(
                         ),
                         20,
                     ),
+                    facet_row_spacing=max((0.08 - (0.00071 * len(parameters))), 0.03),
```
asnyv, line 209: This feels a bit hacky. But yes, I know from experience that this scaling is hard in Plotly / Dash. 
Have you tested on different screen sizes and resolutions?

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
```diff
@@ -206,39 +206,94 @@ def make_grouped_plot(
...
+                    custom_data=["PARAMETER"],
                 )
                 .update_xaxes(
                     matches=None,
                     fixedrange=True,
                     title=None,
-                    showticklabels=(len(parameters) < 20),
+                    showticklabels=len(parameters) <= 100,
+                    tickangle=0,
+                    tickfont_size=max((18 - (0.4 * len(parameters))), 8),
```
asnyv, line 220: Similar thing here, not sure how it works on different screen sizes and resolutions?

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
```diff
@@ -206,39 +206,94 @@ def make_grouped_plot(
...
                         points=False,
                     )
                 )
                 .for_each_annotation(
                     lambda a: a.update(
-                        hovertext=a.text.split("=")[-1],
-                        text=(a.text.split("=")[-1]) if len(parameters) < 40 else "",
+                        text=(a.text.split("=")[-1]),
+                        visible=len(parameters) <= 42,
+                        font_size=max((18 - (0.4 * len(parameters))), 8),
```
asnyv, line 238: Same question on scaling for different screens/windows sizes. 

Also: expression for font_size is the same as above for tickfont_size. Could consider to create a small function to share the logic.

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
```diff
@@ -206,39 +206,94 @@ def make_grouped_plot(
...
+                    text=(
+                        f"<b>{parameter}</b><br>"
+                        + "<br>".join(
+                            f"<b>{ens}:</b><br>"
+                            "Avg: "
+                            f"{self.get_stat_value(parameter, ens, stat_column='Avg'):{form}}<br>"
+                            "Std: "
+                            f"{self.get_stat_value(parameter, ens, stat_column='Stddev'):{form}}"
+                            for ens in ensembles
+                        )
```
asnyv, line 277: Feels unnatural to have one box with hoverinfo per ensemble, but still show data for both in each hoverinfo box (the boxes are equal.  Would rather view only the data for the trace (ensemble) you hover.

webviz_subsurface/plugins/_parameter_analysis/views/selector_view.py
```diff
@@ -304,11 +305,9 @@ def filter_vector_selector(
     )
 
 
-def color_selector(parent, tab: str, px_colors: dict = None, height=None):
-    custom_colors = {
-        "theme": parent.theme.plotly_theme.get("layout", {}).get("colorway", {})[1:12]
-    }
-
+def color_selector(
+    parent, tab: str, colors: list = None, bargap: float = None, height: float = None
```
asnyv, line 309: Might have to
`from typing import Optional`

Mismatch between type hint and a default `None` value are recurring a few other places as well.
`some_variable: some_type = None`
are recurring quite a few places.
```suggestion
    parent, tab: str, colors: Optional[list] = None, bargap: Optional[float] = None, height: Optional[float] = None
```

asnyv (changes requested):
Hi! 
Sorry for late reply 😊 Wrote a few comments in the code. 

Some additional notes: 

- I am _slightly_ worried that we are diverging from the concept of themed styling and coloring, and more towards hardcoded colors, and various other colormaps. I am of course aware that the Equinor theme used internally is not ideal, as the contrasts are often too weak, but "solving" that plugin by plugin is probably not the right way to do it. Making plugin specific coloring will make a later change much more demanding, considering the increasing number of plugins in this repository.
- This plugin is currently in the new list of [ignored plugins for mypy check in CI](url). Not sure how much work there is to make MyPy happy so that we can get it into the regular checks. Consider to give it a go if you have time 😊 As I mention one place, I would at least expect some `MyPy` errors on `some_variable: some_type = None` where you should have  `some_variable: Optional[some_type] = None`. In addition; at least the recurring `parent` input  misses type hinting + missing type hinted return values.

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
tnatt: Yes, agree, this is a bit hacky... but needed when you have many parameters plotted at the same time (otherwise the plots are squeezed and the spacing around them are large). 
From Plotly's documentation they state that this is in reference to the paper coordinates of a plot, so it should work on different screen resolutions. 

Ideally I would like to have this number adjusted by Plotly, they state themselves that it should be modified based on how many plots a user have: 

> The defaults work well with 1-4 rows or columns at the default figure size with the default font size, but need to be reduced to around 0.01 for very large figures or figures with many rows or columns. Conversely, if activating tick labels on all facets, the spacing will need to be increased.

But I guess for now this is the solution at hand.

webviz_subsurface/plugins/_parameter_analysis/views/selector_view.py
tnatt: I wasn't aware of this Optional, thanks 👍

webviz_subsurface/plugins/_parameter_analysis/controllers/parameter_response_controller.py
```diff
@@ -448,7 +446,7 @@ def set_real_color(df_norm, real_no: str):
     Midpoint for the colorscale is set on the average value
     """
     red = "rgba(255,18,67, 1)"
-    mid_color = "rgba(220,220,220,1"
+    mid_color = "rgba(220,220,220,1)"
```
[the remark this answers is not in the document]
tnatt: Well spotted 😉

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+###Added
+- [#509](https://github.com/equinor/webviz-subsurface/pull/509) - Added descriptive hoverinfo to  `ParameterAnalysis`. Average and standard deviation of parameter value
+for each ensemble shown on mouse hover over figure.
```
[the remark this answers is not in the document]
tnatt: Agree, I'll add a sentence regarding the dynamic sizing of titles and subplot layout!

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
tnatt: Actually here I disagree 😉 The point of this parameter plotting is to be able to look at parameter updates from AHM. Having all the information in one hover box makes it easier to compare values from two iterations, instead of having to move back and forth between the traces.

In addition, when the distributions are similar, the boxes will overlap and it would be difficult to hover on the box in the back.

image

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
tnatt: I agree with the small function, perhaps this should be part of a generalized utility library?

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
tnatt: Works fine on laptop and big monitor 👍

HansKallekleiv (approved):
LGTM 👍

asnyv (approved):
👍😊

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
asnyv: I see your point, but is it then necessary with one box per trace, and not just one for the plot? 🤔 no problem to leave it as is for now

webviz_subsurface/plugins/_parameter_analysis/models/parameters_model.py
asnyv: Maybe something similar, could be soemthing we can consider at some point at least 😊 Not sure how well the equation would fit for a general plugin 🤔

PR #529, opened 2020-12-11
Added support for PVDO and PVTG.

rubenthoms:
First approach.

### DIFF
--- a/webviz_subsurface/_datainput/opm_init_io.py
+++ b/webviz_subsurface/_datainput/opm_init_io.py
@@ -116,18 +116,16 @@ class PvxOBase:
 
     # pylint: disable=R0201
     def formation_volume_factor(
-        self, gas_oil_ratio: List[float], pressure: List[float]
+        self, set_value: List[float], pressure: List[float]
     ) -> List[float]:
         """
         Does only return an empty list for now.
         """
-        if len(gas_oil_ratio) != len(pressure):
+        if len(set_value) != len(pressure):
             raise ValueError("rs and po arguments must be of same size")
         return []
 
-    def viscosity(
-        self, gas_oil_ratio: List[float], pressure: List[float]
-    ) -> List[float]:
+    def viscosity(self, set_value: List[float], pressure: List[float]) -> List[float]:
         pass
 
 
--- a/webviz_subsurface/_datainput/opm_init_io.py
+++ b/webviz_subsurface/_datainput/opm_init_io.py
@@ -116,16 +116,18 @@ class PvxOBase:
 
     # pylint: disable=R0201
     def formation_volume_factor(
-        self, set_value: List[float], pressure: List[float]
+        self, gas_oil_ratio: List[float], pressure: List[float]
     ) -> List[float]:
         """
         Does only return an empty list for now.
         """
-        if len(set_value) != len(pressure):
-            raise ValueError("rs and po arguments must be of same size")
+        if len(gas_oil_ratio) != len(pressure):
+            raise ValueError("Rs / Rv and pressure arguments must be of same size")
         return []
 
-    def viscosity(self, set_value: List[float], pressure: List[float]) -> List[float]:
+    def viscosity(
+        self, gas_oil_ratio: List[float], pressure: List[float]
+    ) -> List[float]:
         pass
 
 
@@ -223,6 +225,7 @@ class Oil(Implementation):
 
             # PKey  Inner   C0  C1          C2          C3
             # Rs    Po      1/B 1/(B*mu)    d(1/B)/dPo  d(1/(B*mu))/dPo
+            # :     :       :               :           :
 
             for index_primary in range(0, raw.num_primary):
                 if self.entry_valid(raw.primary_key[index_primary]):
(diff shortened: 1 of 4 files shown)


webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -116,25 +116,27 @@ def get_values(self) -> List[VariateAndValues]:
...
     def formation_volume_factor(
-        self, gas_oil_ratio: List[float], pressure: List[float]
+        self, set_value: List[float], pressure: List[float]
     ) -> List[float]:
         """
         Does only return an empty list for now.
         """
-        if len(gas_oil_ratio) != len(pressure):
+        if len(set_value) != len(pressure):
             raise ValueError("rs and po arguments must be of same size")
```
asnyv, line 125: Suggest to generalize this now
```suggestion
            raise ValueError("Rs / Rv and pressure arguments must be of same size")
```

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -259,6 +260,58 @@ def create_live_oil(self, raw: EclPropertyTableRawData) -> List[PvxOBase]:
...
+
+        column_stride = raw.num_rows * raw.num_tables * raw.num_primary
+        table_stride = raw.num_primary * raw.num_rows
+
+        # pylint: disable=too-many-nested-blocks
+        for index_table in range(0, raw.num_tables):
+            values = []
+
+            # PKey  Inner   C0  C1          C2          C3
+            # Rs    Po      1/B 1/(B*mu)    d(1/B)/dPo  d(1/(B*mu))/dPo
```
asnyv, line 275: I don't know the INIT file that well, but is this comment correct? Is the key Rs for a dead oil? Or is it there anyways, and then just 0?

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -278,6 +331,8 @@ def from_ecl_init_file(ecl_init_file: EclFile) -> Optional["Oil"]:
         if raw.num_tables == 0:
             return None
 
+        # If there is no Rs index, no gas is dissolved in the oil -> dead oil
+        # Otherwise, gas is dissolved and the ratio is given -> live oil
```
asnyv, line 335: Looking at the manual, there are also some other edge cases like PVCO (live oil in compressibility form) and PVCDO (dead oil in compressibility form). I don't think it is much in use, so this is probably ok for now. But I guess we should test these scenarios in the future.

webviz_subsurface/_datainput/opm_init_io.py
```diff
@@ -338,7 +393,7 @@ def create_dry_gas(self, raw: EclPropertyTableRawData) -> List[PvxOBase]:
             values = []
 
             # PKey  Inner   C0  C1          C2          C3
-            # Rs    Po      1/B 1/(B*mu)    d(1/B)/dPo  d(1/(B*mu))/dPo
+            # Pg    Rv      1/B 1/(B*mu)    d(1/B)/dPg  d(1/(B*mu))/dPg
```
asnyv, line 396: Guess this is the same question as for dead oil: do you have a Rv here? But I get a feeling now that it exists, but is constant 0 🙃

webviz_subsurface/_datainput/pvt_data.py
```diff
@@ -159,13 +185,20 @@ def init_to_pvt_data_frame(kwargs: Any) -> pd.DataFrame:
         column_keyword = []
 
         if oil:
+            if len(oil.tables()) > 0 and isinstance(oil.tables()[0], LiveOil):
+                keyword = "PVTO"
+            else:
+                keyword = "PVDO"
```
asnyv, line 191: Might be overkill. But could be a way to catch regression bugs when making changes later?
```suggestion
            if len(oil.tables()) > 0 and isinstance(oil.tables()[0], LiveOil):
                keyword = "PVTO"
            elif len(oil.tables()) > 0 and isinstance(oil.tables()[0], DeadOil):
                keyword = "PVDO"
            else:
                raise SomeSuitableError
```

webviz_subsurface/_datainput/pvt_data.py
```diff
@@ -176,17 +209,21 @@ def init_to_pvt_data_frame(kwargs: Any) -> pd.DataFrame:
                             )
 
         if gas:
+            if len(gas.tables()) > 0 and isinstance(gas.tables()[0], DryGas):
+                keyword = "PVDG"
+            else:
+                keyword = "PVTG"
```
asnyv, line 215: Consider the same as suggested above for LiveOil and DeadOil

webviz_subsurface/_datainput/pvt_data.py
```diff
@@ -37,19 +45,36 @@ def filter_pvt_data_frame(
...
         "KEYWORD",
-        "GOR",
+        "RATIO",
         "PRESSURE",
         "VOLUMEFACTOR",
         "VISCOSITY",
     ]
+
+    if not "RATIO" in data_frame.columns:
+        data_frame["RATIO"] = 1.0
```
asnyv, line 76: What is this for? setting an arbitrary ratio to 1.0 sounds likely to be wrong 🤔

asnyv (changes requested):
Looks good 😊 Added a few comments

asnyv (approved):
LGTM 👍  🎉

PR #531, opened 2020-12-17
Obey column_keys when storing unsmry data with frequency raw

### LINKED ISSUE #528: BhpQC gives MemoryError on fields with e.g. numerical simulation issues
BhpQC can easily gives MemoryError since it uses `raw` as time step. Allow BhpQc to only extract the few vectors it needs for visualization?

This problem is significant on models with e.g. numerical issues (a lot of raw time steps).

asnyv:
Obeys selected `column_keys` when using `time_index` / `sampling` of `raw` (or the equivalent `None`) when loading and storing `UNSMRY` data.

Did some test on larger datasets than our test data, and it was a dramatic improvement. Saw peak memory usage reduced by up to **97%!**

Nevertheless, it is clear that we have memory issues: with this edit on my larger data set, peak memory usage during load and portable build was still more than 3 times higher than peaks when running the portable. 

**Could also be that we should do the same for `daily` data frequency, as that is also quite high when it comes to reservoir simulation?**

---

### Contributor checklist

- [X] :tada: This PR closes #528 

### DIFF
--- a/webviz_subsurface/_models/ensemble_model.py
+++ b/webviz_subsurface/_models/ensemble_model.py
@@ -62,6 +62,24 @@ class EnsembleModel:
         time_index: Optional[Union[list, str]] = None,
         column_keys: Optional[list] = None,
     ) -> pd.DataFrame:
+        # Limit saved columns if time_index = 'raw' or None, as the data density might
+        # be very high, and increases risk of `MemoryError`
+        if time_index == "raw" or time_index is None:
+            self._webviz_store.append(
+                (
+                    self._load_smry,
+                    [
+                        {
+                            "self": self,
+                            "time_index": time_index,
+                            "column_keys": column_keys,
+                        }
+                    ],
+                )
+            )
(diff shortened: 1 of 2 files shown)


webviz_subsurface/_models/ensemble_model.py
```diff
@@ -110,6 +128,13 @@ def _load_smry(
         time_index: Optional[Union[list, str]] = None,
         column_keys: Optional[list] = None,
     ) -> pd.DataFrame:
+        print(time_index)
```
HansKallekleiv, line 131: remove

HansKallekleiv (changes requested):

webviz_subsurface/_models/ensemble_model.py
asnyv: Fixed 👍
How did I miss that 🤔 😣

HansKallekleiv (approved):
LGTM 👍

PR #536, opened 2021-01-05
Fix issue with Pandas 1.2.0 and bump webviz-config dependency for mypy support

HansKallekleiv:
- Remove Pandas 0.24 from CI
- Bump webviz-config, pandas and pyscal dependency
- Fixed an issue related to new pandas version


asnyv (approved):
LGTM 👍

PR #547, opened 2021-01-24
Use column filter when reading parquet from webviz store

HansKallekleiv:
An attempt to reduce memory consuption of plugins using UNSMRY data.

Instead of loading the whole smry dataframe on plugin initialization load relevant vectors as needed.
When in portable a column filter will be added to the webvizstore loading function so that only the relevant columns are loaded.
For non-portable apps the filter will not be applied.

Relates to https://github.com/equinor/webviz-config/pull/387

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/_models/ensemble_model.py
+++ b/webviz_subsurface/_models/ensemble_model.py
@@ -57,10 +57,51 @@ class EnsembleModel:
         )
         return self._load_parameters()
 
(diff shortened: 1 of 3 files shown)


anders-kiaer:
Superseded by #721?

Issue #545, opened 2021-01-21
Shared datasets in memory between plugins (especially summary data)


asnyv:
Shared datasets are now attributes to individual plugins. Especially with large summary dataframes, this becomes an issue. 
Each plugin that uses the data keeps the full dataframe in memory, giving us a lot of memory overhead.

An example:
https://github.com/equinor/webviz-subsurface/blob/fdc63ed99a6bef2b185080809eb6799bada6fa3c/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py#L175 

We should therefore try to somehow share read only dataframes between the plugins to keep the memory consumption as small as possible.

It is likely that we then define settings like sampling, column_filters and etc in the shared settings, and leave the individual plugin settings to only handle plugin specific settings.

It probably requires a substantial amount of refactoring, but the solution we have today doesn't fly for larger datasets.
Also: seems likely that initialization / build becomes faster when we don't do this process for the initialization of each plugin.

PR #557, opened 2021-02-01
Fix numpy type hints + CI

asnyv:
---

### Contributor checklist

- [X] :tada: This PR closes #558.
- [X] :scroll: I have broken down my PR into the following tasks:
   - [X] Unrelated fix in GitHub Actions workflow (env cleaning section), ref: https://github.com/actions/virtual-environments/issues/1057
   - [X] Fix numpy type hints
- [ ] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.
- [ ] :book: I have considered adding a new entry in `CHANGELOG.md`, and added it if should be communicated there.

PR #563, opened 2021-02-10
Avoid mutating and duplicating smry data in SimulationTimeSeriesModels

sigurdp:
This PR implements changes to `SimulationTimeSeriesModel` for both the `ParameterAnalysis` and `PropertyStatistics` plugins. Note that these two plugins do not currently share the model. There exists two relatively similar models.

This PR is related to issue https://github.com/equinor/webviz-subsurface/issues/545 as it resolves some of the underlying issues that causes mutation and/or duplication smry dataframes.

**For `SimulationTimeSeriesModel` in `ParameterAnalysis`:**
Refactored the model to avoid mutating the input smry dataframe and to avoid memory duplication. Removed unused code in preparation for a redesign of the public interface and subsequent consolidation with "near copy" of SimulationTimeSeriesModel that resides within PropertyStatistics.
For now, the part of the public interface that is actually in use has been kept unchanged. Note that previously this model dropped the last date if it was empty. This is no longer the case

**For `SimulationTimeSeriesModel` in `PropertyStatistics`:**
Refactored the model to avoid mutating the input smry dataframe (date to str conversion)

Due to unrelated issues also removed uninstall of `adoptopenjdk*` package for github action as well as bumping required version of `pyscal` to 0.7.5

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/.github/workflows/subsurface.yml
+++ b/.github/workflows/subsurface.yml
@@ -30,7 +30,7 @@ jobs:
       run: |
         # https://github.com/actions/virtual-environments/issues/751
         # https://github.com/actions/virtual-environments/issues/709
-        sudo apt-get purge p7zip* yarn ruby-full adoptopenjdk* ghc* php7*
+        sudo apt-get purge p7zip* yarn ruby-full ghc* php7*
         sudo apt-get autoremove
         sudo apt-get clean
         df -h
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setup(
         # "ecl2df>=0.6.1; sys_platform=='linux'",
         "fmu-ensemble>=1.2.3",
         # "opm>=2020.10.1; sys_platform=='linux'",
-        "pandas>=1.1.5",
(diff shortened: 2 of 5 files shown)


HansKallekleiv (approved):
👍

asnyv:
Think you need to pin pandas to `<= 1.2.1` @sigurdp. Only `<= 1.1.5` that has `py3.6 `support

sigurdp:
@asnyv Will try that change, but strangely the build seems to fail on 3.8 as well. Any ideas?

asnyv:
Hmm, looks like "not our fault" @sigurdp. Same thing fails for all versions now... It is probably related to this: https://github.com/equinor/webviz-subsurface/pull/557/files#r568617848

But why the "fix" now fails... 🤔

asnyv:
@sigurdp Could maybe be: https://github.com/actions/virtual-environments/issues/2446
8th of Feb + 3-4 days fits well

Issue #444, opened 2020-09-22
Disable min-max shade


anders-kiaer:
A user has requested possibility of turning of min-max in e.g. `ReservoirSimulationTimeSeries` (ensemble has extreme values - sometimes easier to compare using p10-p90 as boundaries).

asnyv:
Extended requests: 

- No shade (only lines): Easier to separate when plotting many cases, as the colors don't blend. 

- Possibility to select which active (e.g. min, max, mean, p10, p50, p90). With the current shading, that is a bit tricky. Easy for the lines in the first bullet point, so think might be ok to just have it there.

anders-kiaer:
Considering this closed by #612.

Issue #226, opened 2020-02-10
Default choice for clickData based plots


asnyv:
Include a default initial choice for plots that are activated by clickData to avoid empty figures.
 
Examples:
- Choose best correlated value as default from tornado in plugins/_parameter_correlation.py

- Choose last timestep as default for tornado plot in plugins/_reservoir_simulation_timeseries_onebyone.py

-Choose last timestep  (or mid to avoid all zeros in example?) as default for histogram in plugins/_reservoir_simulation_timeseries.py

Issue #214, opened 2020-01-31
Feature request - simulation time series


sge-gatsby:
I have spent only very little time with webviz,  but as user I would like to:

-	Be possible to include the delta of the new well (even if it is not defined in both ensembles…, I guess if missing from one ensemble could be assumed that all parameters in that ensemble are zero). 
-	I would like to see more deltas included in one plot (i.e,. to compare different deltas between several ensembles…, or maybe in the plots below. Here I'm thinking of comparing (for example) the same reference case but for 2 IOR projects (i.e. 2 different well locations), but also different reference cases and their corresponding IORs...
-	I would also like to switch between possibility of using statistics (only) and the whole ensemble (including statistics), or only ensembles without statistics, or only some chosen statistics (i.e. only P10s).

That’s all for the time being! 😊😊😊

jorgenherje:
PR https://github.com/equinor/webviz-subsurface/pull/867 closes request for multiple deltas in one plot. The statistics options are already handled in PR. https://github.com/equinor/webviz-subsurface/pull/867

Should this issue be closed @anders-kiaer ?

anders-kiaer:
I think we can close this now, after #867. There might be some details remaining, but if that is the case let us specify them in a stand-alone issue highlighting those.

PR #612, opened 2021-04-23
Plot options ReservoirSimulationTimeSeries

### LINKED ISSUE #403: Consider to remove history trace from delta ensemble in ReservoirSimulationTimeSeries
Delta of history trace is visible, can consider to remove it. Differences in history input between ensembles is often 0, unless there are changes in history input. Differences could e.g. be caused by reallocation of history between when the ensembles were run, allocation uncertainty included in assisted history matching, and different length of history (e.g. when comparing an updated model with an older one).
image

### LINKED ISSUE #496: Histogram traces in ReservoirSimulationTimeSeries not toggled by clicking legend
Likely due to missing legendgroup in [add_histogram_traces](https://github.com/equinor/webviz-subsurface/blob/f9062a639974518c4e51c726b57a5c0111810bb9/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py#L1039-L1051)

Example: iter-2 only toggled in fan chart
image

### LINKED ISSUE #503: Option of removing fill in fan charts
The fill makes it hard to compare more than ~2-3 ensembles (or even just 2 if you are unlucky on coloring)

An option to turn off the fill of fan charts. The individual lines for one color may then all be the same color (for easy comparison), maybe different line style for percentiles?

How it looks now:
image

### LINKED ISSUE #504: History vectors are plotted into prediction
Eclipse continues to write to history vectors even after end of e.g. WCONHIST. Looks strange in prediction to have a history line.

Not have a history line in predictions

**Describe alternatives you've considered**
Should be possible to either/or: 
a) remove history line
b) cut history trace at current date (as it is impossible that history is longer)

**Additional context**
If you have uncertainty in history input, the current history trace is not representative (as it only plots the first realization). Plotting history as an ensemble is probably overkill, so could be nice to be able to turn it off also in history phase.

asnyv:
Modifications to ReservoirSimulationTimeseries:

- Allow statistical lines without fill (including new P50/median). (Easier to read, avoids overlapping color fills)
  - Can choose which statistics to include (Mean, P10, P50, P90, Max, Min)
- Option to remove history trace (and cut history trace at current date, as longer is not realistic even though Eclipse creates it)
- Add `legendgroup` to histogram traces to make them "toggleable"
- Allow histogram together with individual realizations
- Remove history traces in delta plots

---

Technically breaking: `statistics_hist` no longer valid config option (but doubt it is used at all). The default `statistics` now refers to statistical lines, whilst `fanchart` refers to the old default. May consider to change the default to `fanchart`, but I think at least the naming is most descriptive this way.

### Contributor checklist

- [X] :tada: This PR closes #503, closes #504, closes #496, closes #403, relates to (might close) #444, relates to (might close) #226, relates to (will not close) #214. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [UNRELEASED] - YYYY-MM-DD
+### Changed
+- [#612](https://github.com/equinor/webviz-subsurface/pull/612) - New features in ReservoirSimulationTimeSeries: Statistical lines, option to remove history trace, histogram available when plotting individual realizations.
 
 ## [0.2.0] - 2021-03-28
-### Changed
 - [#604](https://github.com/equinor/webviz-subsurface/pull/604) - Consolidates surface loading and statistical calculation of surfaces by introducing a shared
 SurfaceSetModel. Refactored SurfaceViewerFMU to use SurfaceSetModel.
 - [#586](https://github.com/equinor/webviz-subsurface/pull/586) - Added phase ratio vs pressure and density vs pressure plots. Added unit and density functions to PVT library. Refactored code and added checklist for plots to be viewed in PVT plot plugin. Improved the layout.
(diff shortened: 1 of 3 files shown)


HansKallekleiv (approved):
LGTM 👍

Issue #617, opened 2021-04-27
Deprecate cross-section plugins


HansKallekleiv:
The `StructuralUncertainty` plugin replaces
- `HorizonUncertaintyViewer`
- `WellCrossSection`
- `WellCrossSectionFMU`

- [ ] Add deprecation warning to these plugins
- [ ] Remove the plugins

anders-kiaer:
Closed in #618.

PR #618, opened 2021-04-27
Add deprecation warning for cross-section plugins

HansKallekleiv:

---

### Contributor checklist

- [x] :tada: This PR relates to #617 
   - [x] Add deprecation warning to `HorizonUncertaintyViewer`, `WellCrossSection`, `WellCrossSectionFMU`.
   - [x] Update webviz-config dependency

### DIFF
--- a/webviz_subsurface/plugins/_horizon_uncertainty_viewer/horizon_uncertainty_viewer.py
+++ b/webviz_subsurface/plugins/_horizon_uncertainty_viewer/horizon_uncertainty_viewer.py
@@ -21,6 +21,7 @@ from webviz_config.webviz_assets import WEBVIZ_ASSETS
 from webviz_config import WebvizPluginABC
 from webviz_config import WebvizSettings
 from webviz_config.webviz_store import webvizstore
+from webviz_config.deprecation_decorators import deprecated_plugin
 
 import webviz_subsurface
 from webviz_subsurface._models import SurfaceLeafletModel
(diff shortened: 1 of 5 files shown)


webviz_subsurface/plugins/_horizon_uncertainty_viewer/horizon_uncertainty_viewer.py
```diff
@@ -29,6 +30,10 @@
 from ._huv_table import FilterTable
 
 
+@deprecated_plugin(
+    "This plugin will soon be deprecated. "
```
anders-kiaer, line 34:
```suggestion
    "The HorizonUncertaintyViewer plugin is deprecated. "
```
It is already (from now) deprecated? Similar for the other two deprecations.

webviz_subsurface/plugins/_well_cross_section.py
```diff
@@ -19,6 +20,10 @@
 from .._datainput.surface import load_surface
 
 
+@deprecated_plugin(  # type: ignore
+    "This plugin will soon be deprecated. "
+    "Relevant functionality is implemented in the StructuralUncertainty plugin."
+)
```
anders-kiaer, line 26: I think `@deprecated_plugin` takes in two arguments (one short string + one longer explanation). Here it is one string only? Maybe that is also why mypy complains?

https://github.com/equinor/webviz-config/blob/master/CONTRIBUTING.md#deprecate-plugins-or-arguments

anders-kiaer (changes requested):
LGTM. A couple of minor comments only - then we can merge this.

anders-kiaer (approved):

PR #635, opened 2021-05-05
Changed layout: show plots on top of each other instead of side by side

rnyb:
Changed layout: show plots on top of each other instead of side by side. This should make it easier to read plots with "normal" window width.

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/plugins/_disk_usage.py
+++ b/webviz_subsurface/plugins/_disk_usage.py
@@ -58,16 +58,8 @@ class DiskUsage(WebvizPluginABC):
                     f"Disk usage on {self.scratch_dir} per user as of {self.date}",
                     style={"text-align": "center"},
                 ),
-                wcc.FlexBox(
-                    children=[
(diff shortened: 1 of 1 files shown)


asnyv:
@rnyb We have tested both before, and landed on side by side. Of course open to reevaluate that decision. Why do you want it changed? 🙂

Old arguments to have them side by side: It is easier to avoid scrollbars by having them side by side, it is more natural to "read"/compare sideways between the plots, and with the flexbox they should get stacked on top of each other if the window is too narrow.

anders-kiaer:
This PR discussion has gone stale. We can easily reopen if needed.

rnyb:
Sorry about that, me and @asnyv  had the discussion on the side of github, but should have posted some notes here as well. Anyways, it's fine to close as we concluded to keep the flexbox, however it might still be worth looking into if we should add a minWidth to the plots (or similar)

Issue #633, opened 2021-05-03
In-place volumes plugin improvements


tnatt:
There is a need to improve the existing volumetrics plugins. The plotting options are limited, and there is no easy source comparison. Often other tools (e.g. Spotfire) are used to present the data instead. 

Desired functionality:
- Today there is two volumetrics plugins in webviz-subsurface, InplaceVolumesOneByOne and InplaceVolumes, it would be good if the new volumetrics plugin could cover functionalities from both. 
- Flexible display of volumes - support subplots and different plot types
- Computation of NTG, poro and sw from the data 
- Easy comparison of sources (model updates or geo vs eclipse volumes) - e.g. compute diff columns
- QC functionality for fast identification of volume changes and causes (bulk/sw/poro)
- Option to split tornado plots in bulk vs inplace sensitivities

Existing issues related to inplace volumes: #433, #388, #296, #236, #100, #493

![image](https://user-images.githubusercontent.com/61694854/116870825-86272300-ac13-11eb-962b-3fb711e1ee6b.png)

PR #641, opened 2021-05-20
Volumetric analysis plugin

tnatt:

---

### Contributor checklist

- [ ] :tada: This PR relates to #633 .
   - [ ] Task 1
   - [ ] Task 2


webviz_subsurface/plugins/_inplace_volumes.py
```diff
@@ -20,7 +20,7 @@
 from .._abbreviations.number_formatting import table_statistics_base
 
 
-class InplaceVolumes(WebvizPluginABC):
+class InplaceVolumesOld(WebvizPluginABC):
```
anders-kiaer, line 23: Something to consider (not sure what is best):

Unless the new `InplaceVolumes` is a drop-in-replacement of the old one (i.e. takes the same argument names), you can also consider deprecating the existing plugin (`webviz-config` recently got functionality for this: https://github.com/equinor/webviz-config/blob/master/CONTRIBUTING.md#deprecate-plugins-or-arguments), and then try (if possible) to find a new name for the new plugin.

Benefit: Existing webviz configurations/setups will continue to exist, and users are notified that they should use something else. Downside: Need a new name for the new plugin, and the transition might take a bit longer (no there is no hard error :laughing: ).

tnatt: Yes, this was the quickest solution for me to get the new plugin running locally while working 😉 I agree I like the new deprecation system, and we can come up with a new clever name for this one!

HansKallekleiv:
Superseded by #648

PR #642, opened 2021-05-20
Including stratigraphy in the well completion plugin

lindjoha:
The main change in this PR is the possibility to pass a json file with a stratigraphic tree to the plugin. The stratigraphy will be combined with the zone_layer mapping file (lyr-file) to generate the final stratigraphy that is passed to the wsc component.

There is also new logic around colors. Colors can be supplied through both the stratigraphy file and the lyr file. If colors is passed in both files, the stratigraphy has priority. If no colors is passed, the theme colors will be used. The parsing of the lyr files with colors is depending on new functionality in ecl2df.

There is a file with tests around the logic of the above.

There is also some new functionality around the kh units. If now kh unit is supplied, the plugin will look in the eclipse DATA file for the unit system, and choose the correct unit according to that.

### DIFF
--- a/webviz_subsurface/_datainput/well_completions.py
+++ b/webviz_subsurface/_datainput/well_completions.py
@@ -6,6 +6,7 @@ import glob
 
 from ecl2df import common
 
+
 def read_zone_layer_mapping(
     ensemble_path: str, zone_layer_mapping_file: str
 ) -> Optional[Dict[int, str]]:
@@ -17,7 +18,8 @@ def read_zone_layer_mapping(
         zonelist = common.parse_lyrfile(filename=filename)
         layer_zone_mapping = common.convert_lyrlist_to_zonemap(zonelist)
         zone_color_mapping = {
-            zonedict["name"]:zonedict["color"] for zonedict in zonelist
+            zonedict["name"]: zonedict["color"]
+            for zonedict in zonelist
             if "color" in zonedict and re.match("^#([A-Fa-f0-9]{6})", zonedict["color"])
         }
         return layer_zone_mapping, zone_color_mapping
@@ -67,6 +69,7 @@ def read_well_attributes(
             for well_data in file_content["wells"]
         }
     return None
+
 
 def read_stratigraphy(
     ensemble_path: str, stratigraphy_file: str
--- a/webviz_subsurface/_datainput/well_completions.py
+++ b/webviz_subsurface/_datainput/well_completions.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict
+from typing import Optional, Dict, List
 import json
 import re
 from pathlib import Path
@@ -66,4 +66,14 @@ def read_well_attributes(
             well_data["alias"]["eclipse"]: well_data["attributes"]
             for well_data in file_content["wells"]
         }
+    return None
+
+def read_stratigraphy(
+    ensemble_path: str, stratigraphy_file: str
+) -> Optional[List[Dict]]:
+    """Searches for a stratigraphy json file on the scratch disk. \
+    If a file is found the content is returned as a list of dicts.
+    """
+    for filename in glob.glob(f"{ensemble_path}/{stratigraphy_file}"):
+        return json.loads(Path(filename).read_text())
     return None
--- a/webviz_subsurface/_datainput/well_completions.py
+++ b/webviz_subsurface/_datainput/well_completions.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, List
+from typing import Optional, Dict, List, Tuple
 import json
 import re
 from pathlib import Path
@@ -9,7 +9,7 @@ from ecl2df import common
 
 def read_zone_layer_mapping(
     ensemble_path: str, zone_layer_mapping_file: str
-) -> Optional[Dict[int, str]]:
+) -> Tuple[Optional[Dict[int, str]], Optional[Dict[str, str]]]:
     """Searches for a zone layer mapping file (lyr format) on the scratch disk. \
     If one file is found it is parsed using functionality from the ecl2df \
     library.
@@ -79,4 +79,31 @@ def read_stratigraphy(
     """
     for filename in glob.glob(f"{ensemble_path}/{stratigraphy_file}"):
         return json.loads(Path(filename).read_text())
+    return None
+
+
+def get_ecl_unit_system(ensemble_path: str) -> Optional[str]:
+    """Returns the unit system of an eclipse deck. The options are \
+    METRIC, FIELD, LAB and PVT-M.
+
+    If none of these are found, the function returns None, even though \
+    the default unit system is METRIC. This is because the unit system \
+    keyword could be in an include file.
+    """
+
+    def ecldata_has_unit_system(unit_system: str, lines: list) -> bool:
+        for line in lines:
+            if line.startswith(unit_system):
+                return True
+        return False
+
+    unit_systems = ["METRIC", "FIELD", "LAB", "PVT-M"]
+    for filename in glob.glob(f"{ensemble_path}/eclipse/model/*.DATA"):
+        with open(filename, "r") as handle:
+            ecl_data_lines = handle.readlines()
+
+        for unit_system in unit_systems:
+            if ecldata_has_unit_system(unit_system, ecl_data_lines):
+                return unit_system
+        return None
     return None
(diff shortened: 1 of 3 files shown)


webviz_subsurface/_datainput/well_completions.py
```diff
@@ -1,21 +1,30 @@
...
+    # pylint: disable=no-member
     for filename in glob.glob(f"{ensemble_path}/{zone_layer_mapping_file}"):
-        return ecl2df.EclFiles("").get_zonemap(filename=filename)
-    return None
+        zonelist = common.parse_lyrfile(filename=filename)
+        layer_zone_mapping = common.convert_lyrlist_to_zonemap(zonelist)
+        zone_color_mapping = {
+            zonedict["name"]: zonedict["color"]
+            for zonedict in zonelist
+            if "color" in zonedict and re.match("^#([A-Fa-f0-9]{6})", zonedict["color"])
```
anders-kiaer, line 24: Do we want to (silently) filter out colors here? Add a `logging.WARNING`?

webviz_subsurface/_datainput/well_completions.py
```diff
@@ -61,3 +70,41 @@ def read_well_attributes(
...
+    If none of these are found, the function returns None, even though \
+    the default unit system is METRIC. This is because the unit system \
+    keyword could be in an include file.
+    """
+
+    def ecldata_has_unit_system(unit_system: str, lines: list) -> bool:
+        for line in lines:
+            if line.startswith(unit_system):
+                return True
+        return False
```
anders-kiaer, line 99:
```suggestion
        return any(line.startswith(unit_system) for line in lines)
```
Maybe the function is not needed if can be this short and is used only once below?

webviz_subsurface/plugins/_well_completions.py
```diff
@@ -30,8 +35,9 @@ class WellCompletions(WebvizPluginABC):
     * **`ensembles`:** Which ensembles in `shared_settings` to visualize.
     * **`compdat_file`:** csvfile with compdat data per realization
     * **`zone_layer_mapping_file`:** Lyr file specifying the zone->layer mapping \
+    * **`stratigraphy_file`:** Json file defining the stratigraphic levels \
     * **`well_attributes_file`:** Json file with categorical well attributes \
```
anders-kiaer, line 39: JSON is all capital, alternatively we can perhaps write "`.json` file" (including the backtick formatting). If so, we could also on the lines above write `.lyr` file, `.csv` file etc.

webviz_subsurface/plugins/_well_completions.py
```diff
@@ -483,10 +570,103 @@ def extract_wells(
...
+    stratigraphy, remaining_valid_zones = filter_valid_nodes(
+        stratigraphy, list(set(layer_zone_mapping.values()))
+    )
+
+    if remaining_valid_zones:
+        raise ValueError(
+            f"""
+            The following zones are defined in the zone->layer mapping, but not in the
+            stratigraphy: {remaining_valid_zones}
+            """
```
anders-kiaer, line 661:
```suggestion
            "The following zones are defined in the zone → layer mapping, "
            f"but not in the stratigraphy: {remaining_valid_zones}"
```
Otherwise the error will be written with a lot of white space between "but not in the" and "stratigraphy [...]".

anders-kiaer (changes requested):
Nice :clap: :rocket: Some minor improvement suggestions below.

webviz_subsurface/_datainput/well_completions.py
lindjoha: yes, good idea. I needed to use logging.warning with lower case it seems.

webviz_subsurface/_datainput/well_completions.py
lindjoha: smart!

webviz_subsurface/_datainput/well_completions.py
```diff
@@ -1,21 +1,47 @@
...
+
+def remove_invalid_colors(zonelist: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """Removes colors in the zonelist from the lyr file that is not 6 digit
+    hexadecimal.
+    """
+    for zonedict in zonelist:
+        if "color" in zonedict and not re.match(
+            "^#([A-Fa-f0-9]{6})", zonedict["color"]
+        ):
+            logging.warning(
```
anders-kiaer, line 20: I think the current line will use the `root` logger. :thinking:  Maybe change to the following (consumers can easier turn of logging on module level):
```suggestion
            logging.getLogger(__name__).warning(
```
Ref.: https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

webviz_subsurface/_datainput/well_completions.py
```diff
@@ -61,3 +87,34 @@ def read_well_attributes(
...
+    If none of these are found, the function returns None, even though \
+    the default unit system is METRIC. This is because the unit system \
+    keyword could be in an include file.
+    """
+    unit_systems = ["METRIC", "FIELD", "LAB", "PVT-M"]
+    for filename in glob.glob(f"{ensemble_path}/eclipse/model/*.DATA"):
+        with open(filename, "r") as handle:
+            ecl_data_lines = handle.readlines()
+
+        for unit_system in unit_systems:
```
anders-kiaer, line 117:
```suggestion
    for filename in glob.glob(f"{ensemble_path}/eclipse/model/*.DATA"):
        with open(filename, "r") as handle:
            ecl_data_lines = handle.readlines()

        for unit_system in ["METRIC", "FIELD", "LAB", "PVT-M"]:
```

webviz_subsurface/plugins/_well_completions.py
```diff
@@ -28,10 +33,11 @@ class WellCompletions(WebvizPluginABC):
     ---
 
     * **`ensembles`:** Which ensembles in `shared_settings` to visualize.
-    * **`compdat_file`:** csvfile with compdat data per realization
-    * **`zone_layer_mapping_file`:** Lyr file specifying the zone->layer mapping \
-    * **`well_attributes_file`:** Json file with categorical well attributes \
-    * **`kh_unit`:** Will normally be mDm
+    * **`compdat_file`:** `.csv` file with compdat data per realization
+    * **`zone_layer_mapping_file`:** `.lyr` file specifying the zone->layer mapping \
```
anders-kiaer, line 37:
```suggestion
    * **`zone_layer_mapping_file`:** `.lyr` file specifying the zone ➔ layer mapping \
```

webviz_subsurface/plugins/_well_completions.py
```diff
@@ -28,10 +33,11 @@ class WellCompletions(WebvizPluginABC):
...
+    * **`stratigraphy_file`:** `.json` file defining the stratigraphic levels \
+    * **`well_attributes_file`:** `.json` file with categorical well attributes \
+    * **`kh_unit`:** Will f.ex be mD·m if unit system is METRIC \
```
anders-kiaer, line 40:
```suggestion
    * **`kh_unit`:** Will e.g. be mD·m if unit system is METRIC \
```

webviz_subsurface/plugins/_well_completions.py
```diff
@@ -283,6 +352,31 @@ def create_ensemble_dataset(
...
 
+def count_leaves(stratigraphy: List[Dict[str, Any]]) -> int:
+    """Counts the number of leaves in the stratigraphy tree"""
+    count = 0
+    for zonedict in stratigraphy:
+        if "subzones" in zonedict:
+            count += count_leaves(zonedict["subzones"])
+        else:
+            count += 1
+    return count
```
anders-kiaer, line 363:
```suggestion
    return sum(
        count_leaves(zonedict["subzones"]) if "subzones" in zonedict else 1
        for zonedict in stratigraphy
    )
```
I haven't tested if this gives the same result, but think it is possible to shorten it to something like this.

anders-kiaer (commented):
I think we are close to be able to merge this :+1: :rocket: 

In addition to the comments above, it might be worth adding a changelog entry in `CHANGELOG.md` and bumping the lower level version constraint on `ecl2df` in `setup.py` (since we depend on new functionality in that dependency). :slightly_smiling_face:

webviz_subsurface/plugins/_well_completions.py
lindjoha: Smart! It works :)

PR #646, opened 2021-05-28
Replaced dropdowns with `VectorSelector`

rubenthoms:
Replaced `dcc.DropDown` with `VectorSelector` for selecting vectors in reservoir simulation timeseries plugin.

### DIFF
--- a/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
+++ b/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
@@ -75,9 +75,7 @@ class ReservoirSimulationTimeSeries(WebvizPluginABC):
 * **`obsfile`**: File with observations to plot together with the relevant time series. \
 (absolute path or relative to config file).
 * **`options`:** Options to initialize plots with:
-    * `vector1` : First vector to display
-    * `vector2` : Second vector to display
-    * `vector3` : Third vector to display
+    * `vectors` : List of vectors to display
     * `visualization` : `realizations`, `statistics` or `fanchart`
     * `date` : Date to show by default in histograms
 * **`line_shape_fallback`:** Fallback interpolation method between points. Vectors identified as \
@@ -257,9 +255,9 @@ folder, to avoid risk of not extracting the right data.
         # Check if initially plotted vectors exist in data, raise ValueError if not.
         missing_vectors = [
             value
-            for key, value in self.plot_options.items()
-            if key in ["vector1", "vector2", "vector3"] and value not in self.smry_cols
-        ]
+            for value in self.plot_options["vectors"]
+            if  value not in self.smry_cols
+        ] if "vectors" in self.plot_options else []
         if missing_vectors:
             raise ValueError(
                 f"Cannot find: {', '.join(missing_vectors)} to plot initially in "
@@ -548,11 +546,12 @@ folder, to avoid risk of not extracting the right data.
                                     "Time series:", style={"font-weight": "bold"}
                                 ),
                                 wsc.VectorSelector(
-                                    id=self.uuid("vector"),
+                                    id=self.uuid("vectors"),
                                     maxNumSelectedNodes=3,
                                     data=self.vector_data,
                                     persistence=True,
                                     persistence_type="session",
+                                    selectedTags=self.plot_options.get("vectors", [self.smry_cols[0]])
                                 ),
                             ],
                         ),
@@ -661,7 +660,7 @@ folder, to avoid risk of not extracting the right data.
         @app.callback(
             Output(self.uuid("graph"), "figure"),
             [
-                Input(self.uuid("vector"), "selectedNodes"),
+                Input(self.uuid("vectors"), "selectedNodes"),
                 Input(self.uuid("ensemble"), "value"),
                 Input(self.uuid("mode"), "value"),
                 Input(self.uuid("base_ens"), "value"),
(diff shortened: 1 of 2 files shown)


HansKallekleiv (changes requested):
Some comments based from testing the plugin interactively on the Reek testdata.
The selector looks good and intuitive to use! A real improvement compared to the dropdown layout.
I encountered two issues that should be fixed:
- Initial vectors specified in the config file is correctly showing in the `VectorSelector`, but is not rendered in the plotly graph.
- Atleast one vector is rendered as subplots 3 times. 
image

Also, the changes here needs an update to the [example config](https://github.com/equinor/webviz-subsurface-testdata/blob/master/webviz_examples/webviz-full-demo.yml#L140)

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -734,17 +761,20 @@ def _update_graph(
...
+                            vector=vec,
+                            interval=cum_interval,
+                            as_date=False,
                         )
                     )
 
             # Make a plotly subplot figure
             fig = make_subplots(
-                rows=len(vectors),
+                rows=len(vectors) if len(vectors) > 0 else 1,
```
anders-kiaer, line 773:
```suggestion
                rows=max(len(vectors), 1),
```

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -734,17 +761,20 @@ def _update_graph(
...
                 cols=2 if "Histogram" in trace_options else 1,
                 shared_xaxes=True,
                 vertical_spacing=0.05,
-                subplot_titles=titles,
+                subplot_titles=titles if len(titles) > 0 else ["No vector selected"],
```
anders-kiaer, line 777:
```suggestion
                subplot_titles=titles if titles else ["No vector selected"],
```

anders-kiaer (commented):
LGTM, I only have a couple of minor comments. Since this touches `AVG_`, `INTVL_` and also a plugin which is used quite a lot, I summon @asnyv for more 👀 before merge.

asnyv:
A wild @asnyv has appeared!

Not sure if you have tested to throw in some csv-input with more or less random column names + checked how so called "UDQs" work? I'll try to do some testing on my side with various inputs to see how things work out. As @anders-kiaer mentions, it is a very frequently used plugin, so wise to spend some time on QC 🧐

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -74,9 +87,7 @@ class ReservoirSimulationTimeSeries(WebvizPluginABC):
 * **`obsfile`**: File with observations to plot together with the relevant time series. \
 (absolute path or relative to config file).
 * **`options`:** Options to initialize plots with:
-    * `vector1` : First vector to display
-    * `vector2` : Second vector to display
-    * `vector3` : Third vector to display
+    * `vectors` : List of vectors to display (max 3 vectors, additional ones are ignored)
```
asnyv, line 90: is this breaking change really necessary? Easy to wrap to existing `vector1`, `vector2`, and `vector3` into a list in the code instead (checking if they are defined of course). I agree that it is cleaner, and probably better, but I think it isn't worth it to break existing setups for this.

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -836,36 +866,34 @@ def _update_graph(
...
+                    # Remove linked x-axis for histograms
+                    if "xaxis2" in fig["layout"]:
+                        fig["layout"]["xaxis2"]["matches"] = None
+                        fig["layout"]["xaxis2"]["showticklabels"] = True
+                    if "xaxis4" in fig["layout"]:
+                        fig["layout"]["xaxis4"]["matches"] = None
+                        fig["layout"]["xaxis4"]["showticklabels"] = True
+                    if "xaxis6" in fig["layout"]:
+                        fig["layout"]["xaxis6"]["matches"] = None
+                        fig["layout"]["xaxis6"]["showticklabels"] = True
```
asnyv, line 889: this extra indent looks funky. Sure this is correct?

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
rubenthoms: The existing setups are still working and we are showing a deprecation warning. I just wanted to remove it from the docs since it is going to be deprecated, i.e. users new to this plugin should rather stick to the new way of defining vectors. Does that make sense?

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
rubenthoms: Hehe, I'll better gonna consult my local pylint installation ;)

rubenthoms: Good catch!

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
asnyv: I see that it is deprecated, I just don't think it is worth it to deprecate it (especially considering we are not making other changes on the input side as far as I know). 
I don't see the point of adding the noise with a deprecation warning (and thus possible questions from users). It is slightly cleaner from the code side of course, but don't underestimate the extra issues you introduce on the user side due to the bunch of existing setups. 

We will probably change other parts of the input for this plugin at a later stage, and considering users don't like to do more changes / updates to their setups than absolutely necessary, I suggest that we delay this deprecation to a time when we make a larger change to the plugin's input.

rubenthoms:
@asnyv @anders-kiaer hmm, I have to admit that I found this plugin to be rather hard to read/understand/maintain. I heard that it was written/extended by several people who were new to the project. If this plugin is one of the most frequently used/important ones, it might be worth it spending some time on refactoring the code/code structure to make it better readable/understandable?

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
rubenthoms: @asnyv I see your point. Well, I am happy with either solution (we would just need to remove the deprecation warning and add back the old doc lines). Your call. What do you think, @anders-kiaer?

asnyv:
@rubenthoms Yes, this plugin has been touched by a lot of hands. It is also among the older ones, so the style of newer plugins probably differentiate a bit. And there are still many features that could/should be included. So I can fully agree that a rewrite would be useful, but before anyone starts a major rewrite, I think we should make a better scope of which features that are needed and how it should look a bit more long term. 

In the meantime: I think as mentioned that we should be careful with changing the user input without good reasons. Better for users to make one (or a few) big changes, than having to frequently update details.

webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
anders-kiaer: Lets revert to current input (`vector{1,2,3}`). I like the suggestion to change to list as input, so created #677 and we can take that one when refactoring the `ReservoirSimulationTimeSeries` plugin.

rubenthoms:
@asnyv @anders-kiaer  Regarding your second point, Asgeir, I created two PRs in both `wcc` (https://github.com/equinor/webviz-core-components/pull/148) and `wsc` (https://github.com/equinor/webviz-subsurface-components/pull/405) which are making the new property `lineBreakAfterTag (boolean)` available (defaults to false).

anders-kiaer (approved):

HansKallekleiv (approved):

PR #658, opened 2021-06-05
Tornado Component

### LINKED ISSUE #236: Tornadoplot: Add tornado table

### LINKED ISSUE #279: Tornado plot tooltips outside figure
Ensure that hover labels are within the boundaries of the tornado plot.

### LINKED ISSUE #643: Tornado chart labels
- Do as ResInsight and put labels on vertixal axis in order to save space?
- Add new label showing "sens case"? Currently people take screenshot + add case name manually (e.g. "Open" or "Closed" if it is a fault sensitivity).

HansKallekleiv:
Improve Tornado visualization

existing:
image

new:
image

### Contributor checklist

- [ ] :tada: This PR closes #643 closes #279  closes #236
   - [ ] Refactor current code into a reusable component
   - [ ] Add table visualization
   - [ ] Improve layout and graph style
   - [ ] Replace hovertext labels with text inside/outside the bars


webviz_subsurface/_assets/js/clientside_functions.js
```diff
@@ -9,5 +9,12 @@ window.dash_clientside = Object.assign({}, window.dash_clientside, {
...
     },
+    get_client_height: function (_triggered) {
+      /*
+          Can be used in a dash callback to get the clientHeight
+          in pixels
+     */
+      return document.documentElement.clientHeight
+    }
   },
 });
```
anders-kiaer, line 20: We don't need to do it in this PR, but if this function is to remain in `webviz-subsurface, we might want to prefix the file with `webviz_subsurfacee`, as well as the function names, in order to prevent namespace collisions with other plugin projects doing similar things.

webviz_subsurface/_components/tornado/_tornado_data.py
```diff
@@ -0,0 +1,279 @@
+from typing import Dict, List, Union
+import pandas as pd
```
anders-kiaer, line 2:
```suggestion
from typing import Dict, List, Union

import pandas as pd
```

anders-kiaer (approved):
👏⭐

webviz_subsurface/_assets/js/clientside_functions.js
HansKallekleiv: Agreed. I can make a separate PR for it.

PR #661, opened 2021-06-07
Rename clientside functions

HansKallekleiv:
Move existing clientside function to a general dash_clientside file to facilitate adding more functions later on.


anders-kiaer (approved):
👍

PR #701, opened 2021-06-30
Preserve uirevision in ReservoirSimulationTimeSeries

asnyv:

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
+++ b/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
@@ -818,17 +818,20 @@ folder, to avoid risk of not extracting the right data.
             )
             fig["layout"] = self.theme.create_themed_layout(fig["layout"])
 
-            if "Histogram" in trace_options:
-                # Remove linked x-axis for histograms
-                if "xaxis2" in fig["layout"]:
-                    fig["layout"]["xaxis2"]["matches"] = None
-                    fig["layout"]["xaxis2"]["showticklabels"] = True
-                if "xaxis4" in fig["layout"]:
-                    fig["layout"]["xaxis4"]["matches"] = None
-                    fig["layout"]["xaxis4"]["showticklabels"] = True
-                if "xaxis6" in fig["layout"]:
-                    fig["layout"]["xaxis6"]["matches"] = None
-                    fig["layout"]["xaxis6"]["showticklabels"] = True
+            # Remove linked x-axis for histograms
+            # Keep uirevision (e.g. zoom) for unchanged data.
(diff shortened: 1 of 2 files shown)


webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -818,17 +818,20 @@ def _update_graph(
...
+                if "Histogram" in trace_options:
+                    c0 = "" if i == 1 else i * 2 - 1
+                    fig["layout"][f"xaxis{i*2}"]["matches"] = None
+                    fig["layout"][f"xaxis{i*2}"]["showticklabels"] = True
+                    fig["layout"][f"xaxis{i*2}"]["uirevision"] = f"{vector}"  # _hist"
+                    fig["layout"][f"yaxis{i*2}"]["uirevision"] = f"{vector}"  # _hist"
+                    fig["layout"][f"yaxis{c0}"]["uirevision"] = f"{vector}"  # _hist"
+                else:
+                    c0 = "" if i == 1 else i
+                    fig["layout"][f"yaxis{c0}"]["uirevision"] = f"{vector}"
```
tnatt, line 834: A suggestion from me before I take vacation today 🌞 
To simplify you could use plotly's update function and reference the subplot columns/rows to achive the same. Just include the lines right before turning the figure into a dictionary.

Something like this:
```
            fig.update_xaxes(uirevision="locked", col=1) # Time axis state kept
            fig.update_xaxes(matches=None, showticklabels=True, col=2)
            for i, vector in enumerate(vectors, start=1):
                fig.update_xaxes(uirevision=vector, row=i, col=2)
                fig.update_yaxes(uirevision=vector, row=i)
```

asnyv: good tip @tnatt 😊

I am not planning to merge this as is. More a proof of concept for the similar time series plugins with minimal rewrite from the old code. Pushed it to a draft PR to get it to my demo, but doesn't seem to work for some reason 🥴
At least we need to test a bit which axes we want to keep and etc (a bit messy with various numbers of subplots, so the vectors switch subplot dependent on how many you choose, if you add the histograms and etc). Think we probably have to live with it working if it ends in the same subplot, but not otherwise...

Also: At least earlier, there was quite a big performance difference between using the plotly graph_objects or plotly express vs handling the dicts directly (which was much more efficient).  Think the difference has decreased lately, and we already use the graph_objects to create the subplots in this plugin, so guess it doesn't matter much here.

Have a nice vacation! ⛱☀

anders-kiaer:
Superseded by #717.

PR #704, opened 2021-07-01
Bugfix TornadoPlot when no sensitivities have impact

### LINKED ISSUE #702: Bug in `TornadoBarChart` if no sensitivities has impact
**Describe the bug**
`TornadoBarChart` fails if no sensitivities has impact for the selected response and "remove sensitivities with no impact" option is actice. 

**Expected behavior**
This should not give error, but just return empty figure instead (see image below)

**Screenshots**
image

tnatt:
PR to fix bug when no sensitivities have impact on a response and the "Remove sensitivities with no impact" is active.
This resulted in error before. Now an empty plot with only the refrence line is drawn.

---

### Contributor checklist

- [x] :tada: This PR closes #702

### DIFF
--- a/webviz_subsurface/_components/tornado/_tornado_bar_chart.py
+++ b/webviz_subsurface/_components/tornado/_tornado_bar_chart.py
@@ -60,7 +60,11 @@ class TornadoBarChart:
                 df["sensname"] = sensname
                 df["case"] = case
                 dfs.append(df)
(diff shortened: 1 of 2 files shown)


anders-kiaer (approved):
LGTM. `pylint` fixed in `master` now.

HansKallekleiv (approved):

PR #712, opened 2021-07-04
Fix callback issue `VolumetricAnalysis`

tnatt:
This PR fixes a callback issue in `VolumetricAnalysis` regarding the `SOURCE` and `ENSEMBLE` filters. These filters were always restored to the first value in the list if it their multi option was False and an user interaction was made. 
Now the value of these filters are only changed when necessary by checking the multi option state upfront. This makes the callback logic cleaner as well.


HansKallekleiv (approved):
lgtm👌

PR #715, opened 2021-07-06
Improved search for unit system

lindjoha:
The get_ecl_unit_system function is improved, using functionality from ecl2df.

The folder name of compdat.csv is changed to share/results/tables/

The WellCompletions plugin code and the well_connection_status ert job are refering to the same file name parameter for the file well_connection_status.parquet.


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ from the water zone if the volumes from the full grid geometry are included, and
 ### Fixed
 - [#666](https://github.com/equinor/webviz-subsurface/pull/666) - Handle operations between surfaces with different topology in `SurfaceViewerFMU`
 - [#675](https://github.com/equinor/webviz-subsurface/pull/675) - Adjust minimum zoom level in surface plugins for visualization of large surfaces.
+- [#715](https://github.com/equinor/webviz-subsurface/pull/715) - After this, the WellCompletion plugin finds the kh unit even if the unit system is in an INCLUDE file. Also, well_connection_status_file refers to the same variable in the plugin and ert job.
```
anders-kiaer, line 26:
```suggestion
- [#715](https://github.com/equinor/webviz-subsurface/pull/715) - After this, the `WellCompletion` plugin finds the kh unit even if the unit system is in an INCLUDE file. Also, `well_connection_status_file` refers to the same variable in the plugin and ert job.
```

webviz_subsurface/plugins/_well_completions.py
```diff
@@ -25,6 +25,8 @@
     read_well_connection_status,
 )
 
+WELL_CONNECTION_STATUS_FILE = "share/results/tables/well_connection_status.parquet"
```
anders-kiaer, line 28:
```suggestion
WELL_CONNECTION_STATUS_FILE = Path() / "share" / "results" / "tables" / "well_connection_status.parquet"
```
and then "downcast" to simple `str` where necessary is perhaps better (compared to converting this variable to `Path` on demand).

anders-kiaer (changes requested):
LGTM!

asnyv (changes requested):
I think this might not be the ideal way to attack it, it feels a bit too hardcoded to fit with a single structure. I suggest that we add this to the ERT job, in which case we could add the unit system as metadata to the parquet files. This ensures that we use the correct datafile than a basic glob guess, and it is also a more long term solution, as I think it would more easily translate to what we need with SUMO. In this case we case we can also do a quick search through the parquets to verify that the unit systems are in fact the same in all ensembles and realizations.

Also some more general comments that surely don't belong only in this PR: 
For the ERT jobs it would be better to switch the input format to the same as most other jobs based on Eclipse files. That means rather than the variable being named `<INPUT>`, it should be `<ECLBASE>`. In that way, the usually defined variable <ECLBASE> in an ERT config will be used by default, and the user could only write e.g. `FORWARD_MODEL WELL_CONNECTION_STATUS` without any further inputs (this is what is used by e.g. internal jobs in `ERT`, `subscript` and etc).

I also noticed that the outputs have been removed from the ERT jobs. I think we have to revert this, as in e.g. reservoir coupled runs you will have multiple files of the same format (one per slave + the master). The well connection files will for example be one per slave (unless you later merge them).

anders-kiaer:
As discussed with @lindjoha and @asnyv:

1) We merge as is on the unit system extraction, and wait for support upstream in `ecl2df` when we then can remove the necessary code in `webviz-subsurface`. See https://github.com/equinor/ecl2df/issues/330.

2) This we should do 👍 With default input we would like to support running the jobs without any arguments.

3) We wait with this as it is a more general discussion/problem (multiple Eclipse simulations per real). I'll create an issue.

anders-kiaer (approved):

asnyv (approved):

PR #719, opened 2021-07-09
Common fan chart input utility

jorgenherje:
Created utility function for providing fanchart traces for plugins

Utility provides list of traces for statistics provided in FanchartData input type. The various statistics can be excluded if not wanted in plot - thereby optional as input.

Mean is made independent of the shades, i.e. the mean is plotted on top without any fill: "tonexty" setting.

---

### Reviewer checklist

- [ ] :scroll: Verify functionality in each plugin:
   - [ ] _reservoir_simulation_timeseries.py
   - [ ] _reservoir_simulation_timeseries_reqional.py
   - [ ] _bhp_qc.py
   - [ ] _relative_permability.py
   - [ ] _rft_plotter/_formation_figure.py

- [ ] Verify disabling/excluding of mean, maximum, minimum, low and high input to FanchartData to see effect of "disabling" statistics. Note fill color in fanchart depends on which statistics are included in trace list (plot).

- [ ] Verify that mean is made independent of the shades (if mean values above high percentile value or below low percentile value) works fine.


webviz_subsurface/_utils/fanchart_plotting.py
```diff
@@ -0,0 +1,203 @@
...
+# pylint: disable=too-many-locals
+def get_fanchart_traces(
+    data: FanchartData,
+    color: str,
+    legend_group: str,
+    line_shape: str = "linear",
+    xaxis: str = "x",
+    yaxis: str = "y",
+    low_percentile_name: str = "P90",
+    high_percentile_name: str = "P10",
```
asnyv, line 80: Consider to drop the default, to make sure that those using the feature actually think about it.
```suggestion
    low_percentile_name: str,
    high_percentile_name: str,
```

asnyv:
Overall it looks good 😊

A few comments:
I think we should include some conditions on the data to ensure that the fancharts make sense (need both bounds for a fan). It could maybe also be an expectation from the plugin author to think about it and ensure that it makes sense if we don't want to implement the check (check is of course safer):
* if you have `maximum` you also must have `minimum` and vica versa. 
* if you have `low` you also must have `high` and vica versa. 


I also suggest that we redefine "mean" to a "free line trace", so that you can plot whatever you want, e.g. mean or a chosen percentile on top of the fan chart (for example P50 (median) might sometimes be more descriptive that mean).

webviz_subsurface/_utils/fanchart_plotting.py
jorgenherje: Handled in refactoring in last commit

jorgenherje:
I've refactored the fanchart plot data, and made min/max and low/high data types (dataclasses). These are set optional, thereby one can ensure that min/max are set as a pair, and the user will not provide min without max. The same yields for low/high.

Names are included into datatypes for low/high and "free line" so the user have to specify which statistical data is provided.

This of course adds some datatypes, but will clarify which statistics are dependent of each other for the fanchart.

asnyv:
Overall looks good to me :) 

Did a search for fanchart in this repo, and found a couple of more occurences of similar plots:
[property statics](https://github.com/equinor/webviz-subsurface/blob/91226ffe14a6c8946db711ecd330301dc71cd3a4/webviz_subsurface/plugins/_property_statistics/models/simulation_timeseries_model.py#L250) and [XSectionFigure](https://github.com/equinor/webviz-subsurface/blob/3653f79e89b855dccca32cf1a22d853919b84902/webviz_subsurface/_datainput/xsection.py#L460)

Seems like the `XSectionFigure` is only used by deprecated plugins, so can probably skip that one, but the first one could be nice to add to this PR :)

jorgenherje:
Regarding [property statics](https://github.com/equinor/webviz-subsurface/blob/91226ffe14a6c8946db711ecd330301dc71cd3a4/webviz_subsurface/plugins/_property_statistics/models/simulation_timeseries_model.py#L250) and [XSectionFigure](https://github.com/equinor/webviz-subsurface/blob/3653f79e89b855dccca32cf1a22d853919b84902/webviz_subsurface/_datainput/xsection.py#L460):

`XSectionFigure` neglected due to deprecated plugins. `Property statistics` seems to be unused code - an issue on code clean-up is created for this: #736

jorgenherje:
I've added the usage of fanchart utility in the [Structural Uncertainty](https://github.com/equinor/webviz-subsurface/tree/master/webviz_subsurface/plugins/_structural_uncertainty) plugin in commit https://github.com/equinor/webviz-subsurface/pull/719/commits/05ad18747ff121a9e059c824d9bb17e498203644

Could you please review this @HansKallekleiv ?

 :scroll: Notes:
- I've utilized the fanchart utility for the "uncertainty envelope" data. 
- As mean is an optional statistical feature in the plugin, I did not add "mean" as a "free line trace" for the fanchart data, as I felt it should be excluded from the plot when not selected. 
- As hovering was not functional for the existing "envelope" data, I added an argument to disable hoverinfo for the fanchart traces. 
- Important to notice that min and max data is duplicated in the plugin as it is used in both fanchart and for statistical data traces (dashed lines). This could be fixed by adding the possibility to set line width and line style for each individual trace in the fanchart. However I did not do this, as I see the fanchart as a single line trace with fading towards the statistical features (min/max and p10/p90 percentiles). Of course this can be up for discussion.

setup.py
```diff
@@ -92,6 +92,8 @@
         "webviz-config>=0.3.1",
         "webviz-subsurface-components>=0.4.3",
         "xtgeo>=2.14",
+        "dataclasses>=0.8; python_version<'3.7'",
+        "types-dataclasses>=0.1.5; python_version<'3.7'",
```
anders-kiaer, line 96: Insert them alphabeticalle?

jorgenherje: Fixed

jorgenherje:
Have Fixed order in setup.py and rebased on equinor/master @anders-kiaer

anders-kiaer (approved):

PR #729, opened 2021-08-14
Layout updates to `ParameterFilter`

tnatt:
PR includes:
- smaller font on header
- prevent having to scroll down to find "apply" and "reset" buttons by inheriting size of parent div instead of setting fixed height 
- Position and color adjustment of "apply" and "reset" button 

Before          |  After
:-------------------------:|:-------------------------:
image  |  image


HansKallekleiv (approved):
Lgtm!

PR #741, opened 2021-08-24
Bugfix for tornados in `VolumetricAnalysis` if missing BULK column 

tnatt:
Bugfix if BULK is missing from input show only custom mode to avoid errors in the tornados.


anders-kiaer:
Is this a bug fix on something previously released? If so, worth a changelog entry?

tnatt:
> Is this a bug fix on something previously released? If so, worth a changelog entry?

I will add a changelog entry 👍 :)

tnatt:
> > Is this a bug fix on something previously released? If so, worth a changelog entry?
> 
> I will add a changelog entry 👍 :)

hmm I mixed up what was released and not.. this is a bug-fix on a not released feature!

anders-kiaer (approved):

PR #744, opened 2021-08-24
Removed warning from well completion documentation

lindjoha:
Removed the warning about ignoring WELOPEN lines that are referring to COMPLUMP numbers. This is now fixed in ecl2df.


HansKallekleiv (changes requested):
Bump the `libecl` requirement in `setup.py`?

anders-kiaer:
> Bump the `ecl2df` requirement in `setup.py`?

I had the same thought, but looks like it is not released yet in `ecl2df` (https://github.com/equinor/ecl2df/pull/333). Maybe it can be considered a "bug fix" in `ecl2df` @lindjoha and therefore we can remove the warning on our side and still not bump `ecl2df`?

lindjoha:
I suppose a new release of ecl2df isn't very far away. This new functionality was intended to go into the september release. So I'm not sure. The plugin is not depending on this functionality of course.

anders-kiaer:
> I suppose a new release of `ecl2df `isn't very far away. This new functionality was intended to go into the september release. So I'm not sure. The plugin is not depending on this functionality of course.

:+1: Let's merge. We will bump the `webviz-subsurface-components`requirement anyway in `setup.py`later when that has an official september release. We could consider doing the same for `ecl2df`.

anders-kiaer (approved):

PR #747, opened 2021-08-30
Filter on OK in `ensemble_table_provider_factory`

### LINKED ISSUE #745: EnsembleTableProvider lacks 'OK'-file filter
**Describe the bug**
EnsembleTableProvider does not filter initialized ensembles on existence of `OK` file. E.g. `LinePlotterFMU` fails if any realization lacks the csv / has failed.

**Expected behavior**
Skip realizations with missing `OK` file

asnyv:
We should agree on how far out in the factory code the filter option should be exposed. 
Right now, I added to only the innermost function wrapping the call to `ScratchEnsemble` and then defaulting the filter to `OK`, but could maybe be use cases where the filter is not wanted, and where the option of no filter (by setting `filter_file = None` ) is useful?

---

### Contributor checklist

- [X] :tada: This PR closes #745 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#717](https://github.com/equinor/webviz-subsurface/pull/717) - Keep zoom state in `ReservoirSimulationTimeseries` (inc `Regional` and `OneByOne`) and `RelativePermeability` plugins using `uirevision`.
 - [#707](https://github.com/equinor/webviz-subsurface/pull/707) - Generalized and improved some plot functions in  `PropertyStatistics`, `ParameterAnalysis` and `VolumetricAnalysis`. Replaced histogram with distribution plot in `PropertyStatistics`.
 
+### Fixed
+- [#747](https://github.com/equinor/webviz-subsurface/pull/747) - Added missing realization filter on `OK` file in `EnsembleTableProviderFactory`.
+
 ## [0.2.4] - 2021-07-13
 
 ### Added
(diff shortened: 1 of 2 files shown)


sigurdp:
Until `filter_file` is exposed in the external API of the factory I think we should just hard-code it inside the `_get_or_create_scratch_ensemble()` function instead of having it as a default parameter.

If we **do** want to have it as a parameter to `_get_or_create_scratch_ensemble()`, the `filter_file` parameter must be incorporated into the generation of the cache `key`.

anders-kiaer:
Discussed with @asnyv: we hardcode `OK` and add a changelog entry. Adding this PR to the September release as a bug fix.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#717](https://github.com/equinor/webviz-subsurface/pull/717) - Keep zoom state in `ReservoirSimulationTimeseries` (inc `Regional` and `OneByOne`) and `RelativePermeability` plugins using `uirevision`.
 - [#707](https://github.com/equinor/webviz-subsurface/pull/707) - Generalized and improved some plot functions in  `PropertyStatistics`, `ParameterAnalysis` and `VolumetricAnalysis`. Replaced histogram with distribution plot in `PropertyStatistics`.
 
+### Fixed
+- [#707](https://github.com/equinor/webviz-subsurface/pull/747) - Added missing realization filter on `OK` file in `EnsembleTableProviderFactory`.
```
anders-kiaer, line 19:
```suggestion
- [#747](https://github.com/equinor/webviz-subsurface/pull/747) - Added missing realization filter on `OK` file in `EnsembleTableProviderFactory`.
```

anders-kiaer (approved):
A typo in changelog. Except that (and a rebase) LGTM! 👍

PR #748, opened 2021-08-30
Bugfix in `VolumetricAnalysis` to trigger callbacks on page refresh

tnatt:
Bugfix `VolumetricAnalysis` to trigger callbacks on a page-refresh. 

- Memory storage using dcc.Store has been added to keep track of which pages has been initially loaded after a page refresh. This info is in turn used to trigger the necessary callbacks for displaying data.
- Also removed two unused dcc.Store variables


anders-kiaer (approved):
Add changelog entry regarding the fix?

tnatt:
> Add changelog entry regarding the fix?

I forgot to mention, it is a bugfix on a bug not yet released. It was introduced in #724. Should I still put in an entry?

anders-kiaer:
> I forgot to mention, it is a bugfix on a bug not yet released. It was introduced in #724. Should I still put in an entry?

Then never mind my comment 🙂 Let's merge.

PR #761, opened 2021-09-12
Store `xtgeo.RegularSurface` as bytestream

HansKallekleiv:
Store xtgeo.RegularSurface as bytestream instead of serializing to json for increased performance.

Note that older portable applications will not work with this change.


anders-kiaer (approved):
👍 🏎️

PR #765, opened 2021-09-13
Use correct inline/xline ranges for axes in ` SegyViewer`  z-slice graph

### LINKED ISSUE #764: Depth slice in the SEG-Y viewer might be displayed wrongly
_Remember that this is an open source project. Don't include e.g. screenshots with data that can't be publicly available._

**Describe the bug**
If the inline and or crossline numbers don't start at zero, the Depth slice will be wrongly positioned

**How to reproduce**
1. Select a seg-y cube where the inline and crossline numbers are offset from zero
2. Display the SEG-Y viewer
3. Notice that the depth slice is starts at 0,0 even if these inlines and crosslines doesn't exist and that the cross that indicates which inlines and crosslines that are being displayed, is positioned in the correct place

**Expected behavior**
The depth slice and the cross should be at the same position and the axis should start at the first inlines and crosslines

**Screenshots**
Screenshots are available, but can't be published

ashildskalnes:
This is a small fix to make sure that the z-slice is displayed at the correct position independent of the range of inlines and crosslines

### Contributor checklist

- [x] :tada: This PR closes #764.
   - [ ] Task 1
   - [ ] Task 2


anders-kiaer:
Nice @ashildskalnes - this looks good to me. Could you add a new entry in `CHANGELOG.md` before we merge this? I.e. add a new `Fixed` section under `[UNRELEASED]` here: https://github.com/equinor/webviz-subsurface/blob/master/CHANGELOG.md#unreleased---yyyy-mm-dd

anders-kiaer (approved):
👍

Issue #693, opened 2021-06-28
Prepare input data to `GroupTree` component


anders-kiaer:
...by using e.g. `ecl2df`, `ecl` etc.

See https://equinor.github.io/webviz-subsurface-components/ for demo of component (choose `GroupTree` in dropdown).

See here for current data format: https://github.com/equinor/webviz-subsurface-components/blob/master/src/demo/example-data/group-tree.json

PR #771, opened 2021-09-20
Grouptree plugin for network vizualisation

lindjoha:
Solves #693. 

Combines summary and gruptree data (exported with `ecl2df`) to generate the data set that is input to the GroupTree component in `webviz-subsurface-components`. 

Includes some selection and filtering options.


webviz_subsurface/plugins/_group_tree/views/filters_view.py
```diff
@@ -0,0 +1,26 @@
+from typing import Callable
+
+import webviz_core_components as wcc
+
+
+def filters_layout(get_uuid: Callable) -> wcc.Selectors:
+    """Descr"""
```
anders-kiaer, line 7: Insert description?

anders-kiaer (commented):
Nice, not able to find much to comment out. Again a new awesome plugin addition, utilizing the `ecl2df` + `webviz-subsurface` + `webviz-subsurface-components` stack. :rocket:

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,212 @@
...
+    * WWIR and/or WGIR for all injectors
+
+    **GRUPTREE input**
+
+    `gruptree_file` is a path to a file stored per realization (e.g. in \
+    `share/results/tables/gruptree.csv"`).
+
+    The `gruptree_file` file can be dumped to disk per realization by a forward model in ERT that
+    wraps the command `ecl2csv compdat input_file -o output_file` (requires that you have `ecl2df`
+    installed).
```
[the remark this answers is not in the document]
berland: Yup, see https://fmu-docs.equinor.com/docs/ert/reference/forward_models.html#ECL2CSV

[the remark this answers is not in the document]
lindjoha: Yes :) I knew that! Will fix

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,215 @@
+from typing import List, Dict, Tuple, Callable
+
+import pandas as pd
+import dash
+import dash_html_components as html
```
anders-kiaer, line 5: We should use the new `from dash import html` 🙂 https://dash.plotly.com/dash-2-0-migration

asnyv: to be explicit:
```suggestion
from dash import Dash, html
```
and further down replace `app: dash.Dash` with `app: Dash` to be consistent with the changes we made in other plugins for `Dash 2.0` 😊
The same import update needs to be made several places

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,215 @@
...
+    ens_name: str, ens_path: str, gruptree_file: str
+) -> pd.DataFrame:
+    """Reads the gruptree file for and ensemble.
+
+    If BRANPROP is found in the KEYWORD column, then GRUPTREE rows
+    are filtered out.
+
+    If the trees are equal in every realization, only one realization is kept.
+    """
+    ens = ScratchEnsemble("ens", ens_path)
```
asnyv, line 184: the ensemble needs to be filtered on `OK` for consistency, but I think it might be better to use the ensembles through the EnsembleSetModel in `self.emodel` should already contain ensembles in the internal property `_ensembles` (should maybe be exposed?), and the filtering is also handled there. 
Maybe better to use that instead of introducing a new place where we initialize ensembles (mainly to keep control of ensemble initializations and avoid the classic "missing OK filter error" that we encounter a bit too often 🙈).

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,215 @@
...
+            i > 0
+            and gruptrees_are_equal
+            and not df_real[compare_columns].equals(df_prev)
+        ):
+            gruptrees_are_equal = False
+        else:
+            df_prev = df_real[compare_columns].copy()
+
+        df_real["REAL"] = row["REAL"]
+        df_all = pd.concat([df_all, df_real])
```
asnyv, line 210: Would guess that it is faster to temporarily store a list of all df's for realizations and then concat all at once, rather than concatenating a realization df with a growing "total df" for each realization (probably not massive data amounts here, so could be that it is negligible, but percentage wise for this step I would think it is significant).

asnyv (commented):
First of all: From what I've seen of demos, it looks very promising 😊

I have though written a few comments based on what I read from the code (haven't tested yet). I think we have briefly discussed most of these cases before, but don't think we've written it down.

In short (in order of what I right now feel is most important, though that feeling might change 😉): 

- My main concern short term is probably with the handling of summary data, as I think we might be stepping into a scenario with duplicated data in memory here 🤔 (though I hope I am wrong 😉)

- Secondly, for some fields I think it will be necessary to be overwrite the currently defined vectors (at least for gas rates) to get physical pipeline rates accounting for e.g. lift gas, and import/consumption. 

- Thirdly, it seems like we are dependent on well data(?), which we at some point should be independent of.

webviz_subsurface/plugins/_group_tree/group_tree.py
lindjoha: OK

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,215 @@
...
+                    "WBHP*",
+                    "WMCTL*",
+                    "WSTAT*",
+                ],
+            )
+        )
+        smry = self.emodel.get_or_load_smry_cached()
+        gruptree = read_gruptree_files(self.ens_paths, self.gruptree_file)
+        smry["DATE"] = pd.to_datetime(smry["DATE"]).dt.date
+        gruptree["DATE"] = pd.to_datetime(gruptree["DATE"]).dt.date
```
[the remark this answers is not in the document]
lindjoha: I have removed the time_index and columns_keys and the filtering to yearly will now be done internally in the plugin.

webviz_subsurface/plugins/_group_tree/controllers/controllers.py
```diff
@@ -0,0 +1,86 @@
...
+        tree_mode_options: List[Dict[str, Any]] = [
+            {
+                "label": "Ensemble mean",
+                "value": "plot_mean",
+            },
+            {
+                "label": "Single realization",
+                "value": "single_real",
+            },
+        ]
```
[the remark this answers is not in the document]
lindjoha: Good ideas. I will make an issue.

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,215 @@
...
+    This plugin needs the following summary vectors to be exported:
+    * FOPR, FWPR, FOPR, FWIR and FGIR
+    * GPR for all group nodes in the network
+    * GOPR, GWPR and GGPR for all group nodes in the production network \
+    (GOPRNB etc for BRANPROP trees)
+    * GGIR and/or GWIR for all group nodes in the injection network
+    * WSTAT, WTHP, WBHP, WMCTL for all wells
+    * WOPR, WWPR, WGPR for all producers
+    * WWIR and/or WGIR for all injectors
+
```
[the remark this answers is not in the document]
lindjoha: Good points! I think we should make an issue (or two) for these.

webviz_subsurface/plugins/_group_tree/group_tree_data.py
```diff
@@ -0,0 +1,478 @@
...
+    # Recursively find well types of all wells connected to the group node
+    # Deduce group node type from well types
+    groupnodes = nodes[nodes["KEYWORD"] != "WELSPECS"]
+    for _, groupnode in groupnodes.iterrows():
+        wells_are_prod, wells_are_inj, wells_are_other = get_leaf_well_types(
+            groupnode["CHILD"], nodes
+        )
+        is_prod_map[groupnode["CHILD"]] = any(wells_are_prod)
+        is_inj_map[groupnode["CHILD"]] = any(wells_are_inj)
+        is_other_map[groupnode["CHILD"]] = any(wells_are_other)
```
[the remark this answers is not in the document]
lindjoha: Yes, I agree. I'll see what I can do with this one or if it'll have to wait for the next iteration

webviz_subsurface/plugins/_group_tree/controllers/controllers.py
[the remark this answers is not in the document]
lindjoha: https://github.com/equinor/webviz-subsurface/issues/785

webviz_subsurface/plugins/_group_tree/group_tree.py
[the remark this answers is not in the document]
lindjoha: Issue on overwriting vectors used: https://github.com/equinor/webviz-subsurface/issues/786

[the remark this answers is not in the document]
lindjoha: Issue on supporting groups as leaf nodes: https://github.com/equinor/webviz-subsurface/issues/787

webviz_subsurface/plugins/_group_tree/group_tree_data.py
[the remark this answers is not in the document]
lindjoha: This is probably necessary for the issue of supporting groups as leaf nodes. So I will add it to that issue: https://github.com/equinor/webviz-subsurface/issues/787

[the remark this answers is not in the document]
lindjoha: But do we have to require that f.ex WWIR and WGIR are exported also for producers? Or can we assume that the well is not an injector if these summary vectors are not there?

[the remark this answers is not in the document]
asnyv: I think so @lindjoha. Hard to be 100% accurate in this case, but think we can leave some responsibility to users. They (hopefully) also now their own production system well enough to notice if a well is wrongly defined

webviz_subsurface/plugins/__init__.py
```diff
@@ -58,3 +58,4 @@
 from ._well_log_viewer import WellLogViewer
 from ._assisted_history_matching_analysis import AssistedHistoryMatchingAnalysis
 from ._well_completions import WellCompletions
+from ._group_tree import GroupTree
```
berland, line 61: (unsorted list of imports. fancy `isort`?)

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,201 @@
...
+            caching_ensemble_set_model_factory.get_or_create_model(
+                ensemble_paths={
+                    ens: webviz_settings.shared_settings["scratch_ensembles"][ens]
+                    for ens in ensembles
+                }
+            )
+        )
+        smry = self.emodel.get_or_load_smry_cached()
+        gruptree = read_gruptree_files(self.ens_paths, self.gruptree_file)
+        smry["DATE"] = pd.to_datetime(smry["DATE"])  # .dt.date
```
berland, line 90: rm the `# .dt.date` and group the lines that manipulates the smry date?

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def set_callbacks(self, app: dash.Dash) -> None:
+        controllers(app=app, get_uuid=self.uuid, grouptreedata=self.grouptreedata)
+
+
+@webvizstore
+def read_gruptree_files(ens_paths: Dict[str, str], gruptree_file: str) -> pd.DataFrame:
+    """Searches for gruptree files on the scratch disk. These
+    files can be exported in the FMU workflow using the ECL2CSV
+    forward job with subcommand gruptree.
```
berland, line 149: forward job -> forward model

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,201 @@
...
+        df_ens = read_ensemble_gruptree(ens_name, ens_path, gruptree_file)
+        df_ens["ENSEMBLE"] = ens_name
+        df = pd.concat([df, df_ens])
+    return df.where(pd.notnull(df), None)
+
+
+def read_ensemble_gruptree(
+    ens_name: str, ens_path: str, gruptree_file: str
+) -> pd.DataFrame:
+    """Reads the gruptree file for and ensemble.
```
berland, line 162: and -> an

webviz_subsurface/plugins/_group_tree/group_tree.py
```diff
@@ -0,0 +1,201 @@
...
+
+    If the trees are equal in every realization, only one realization is kept.
+    """
+    ens = ScratchEnsemble("ens", ens_path)
+    df_files = ens.find_files(gruptree_file)
+
+    if df_files.empty:
+        raise ValueError(f"No gruptree file available for ensemble: {ens_name}")
+
+    # Load all gruptree dataframes and check if they are equal
```
berland, line 175: Should this dataframe comparison be refactored out into its own testable function?

webviz_subsurface/plugins/_group_tree/group_tree_data.py
```diff
@@ -0,0 +1,478 @@
...
+
+class GroupTreeData:
+    """This class holds the summary and gruptree datasets and functionality
+    to combine the data and calculate the GroupTree component input dataset.
+    """
+
+    def __init__(self, smry: pd.DataFrame, gruptree: pd.DataFrame):
+        self.smry = smry
+        self.gruptree = gruptree
+        self.ensembles = gruptree["ENSEMBLE"].unique()
```
berland, line 18: Could it be nice to include typing information here?

ala `self.ensembles : Set[str] = gruptree["ENSEMBLE"].unique()`

webviz_subsurface/plugins/_group_tree/group_tree_data.py
```diff
@@ -0,0 +1,478 @@
...
+        """This method is called when an event is triggered to create a new dataset
+        to the GroupTree plugin. First there is a lot of filtering of the smry and
+        grouptree data, before the filtered data is sent to the function that is
+        actually creating the dataset.
+
+        Returns the group tree data and two lists with dropdown options for what
+        to display on the edges and nodes.
+
+        A sample data set can be found here:
+        https://github.com/equinor/webviz-subsurface-components/blob/master/react/ \
```
berland, line 63: leave on one line and add `'# noqa` after the final `"""` will make copy and paste easier ;)

webviz_subsurface/plugins/__init__.py
lindjoha: hva do you think @anders-kiaer ?

webviz_subsurface/plugins/_group_tree/group_tree_data.py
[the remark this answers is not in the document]
lindjoha: Yes, let's come back to this discussion.

webviz_subsurface/plugins/__init__.py
anders-kiaer: 👍🏻 Leave it as a new issue atm, we can look at isort (+"fix" sorting in other modules as well) in a separate PR

webviz_subsurface/plugins/_group_tree/group_tree.py
lindjoha: I solved this one by importing the `scratch_ensemble` function from `_datatinput/fmu_input`. That function has the OK-filter and is cached so should solve the memory issue.

webviz_subsurface/plugins/__init__.py
lindjoha: https://github.com/equinor/webviz-subsurface/issues/789

webviz_subsurface/plugins/_group_tree/group_tree.py
lindjoha: Yes, good idea. I don't have time to do it now so I will make an issue.

lindjoha: https://github.com/equinor/webviz-subsurface/issues/790

anders-kiaer (approved):

PR #772, opened 2021-09-20
Add script for doing batch conversion of UNSMRY files to .arrow

sigurdp:
This PR adds script , `smry2arrow_batch` for doing batch conversion of UNSMRY files in an ensemble to .arrow files.
Converted smry data in .arrow format for `eclipse/model/SOME.UNSMRY` will be stored in the file `share/results/unsmry/SOME.arrow`

The script utilizes functionality in `ecl2df` for importing UNSMRY files and writing .arrow files.

---

### Contributor checklist

   - [X] Script for doing batch conversion of SMRY files to .arrow
   - [X] ~~Modify existing smry2arrow script to align output paths~~
   - [X] Must agree upon the dictionary key under which to store per vector smry metadata (see https://github.com/equinor/ecl2df/pull/358)


sigurdp:
We should discuss and agree upon the dictionary key under which to store per vector smry metadata.

https://github.com/equinor/webviz-subsurface/blob/7d58b5a11bf8f027c447dd54a1b30f5e2881f65f/webviz_subsurface/ert_jobs/smry2arrow.py#L119

This relates to PR over in https://github.com/equinor/ecl2df/pull/358 which essentially writes the same data and https://github.com/equinor/webviz-subsurface/pull/721 which consumes the metadata.

anders-kiaer:
@berland ⬆️ Could you help taking a look here 👀 and see if you have comments/opinions, especially related to @sigurdp comment above (that will make the transition/move to `ecl2df` easier when webviz use case is matured, i.e. less perturbations of how/where data is stored).

berland:
I think it is possible to reach komodo 2021.11 for the PR in ecl2df for arrow support, including an ert forward model, meaning I whink webviz_subsurface in komodo 2021.11 should aim to use that. We need to agree on the metadata storage format for sure :)

On first look, I am not comfortable with having a script that loops over the directory structure to convert unsmry to arrow, that task should be left with ERT. Is this script some temporary band-aid solution?

anders-kiaer:
> On first look, I am not comfortable with having a script that loops over the directory structure to convert unsmry to arrow, that task should be left with ERT. Is this script some temporary band-aid solution?

Definitely. This is band-aid for ensembles on disk already, for projects wanting to use new plugins not supporting reading from `.UNSMRY` anymore. The ERT job will be the preferred way for the user.

berland:
> Definitely. This is temporary band-aid for ensembles on disk already, for projects wanting to use new plugins not supporting reading from `.UNSMRY` anymore. The ERT job will be the preferred way for the user and in new FMU runs/ensembles.

`$ find . -name *.UNSMRY -exec echo ecl2csv --verbose --arrow {} -o {}.arrow \; | sed 's/UNSMRY.arrow/arrow/g' | sh -x`
🤔

asnyv:
@berland there surely are ways to do this by yourself, and we did consider to try to spread the word of a command like the one you propose. But we decided that if we are to _force_ all users to convert their files to `arrow` (also for old ensembles),  we should do what we can to make the transition as smooth as possible for the average user, and your suggested solution might look a bit "frightening" to many? 

But what if we e.g. write a clear message to terminal when using the band-aid solution, telling the user that this is a limited time offer, and not planned to be the permanent solution, with e.g. links to the (more) future-proof jobs (ert job for new ensembles, ecl2arrow for single cases)? Maybe even go straight to `DeprecationWarning` 😅

webviz_subsurface/ert_jobs/smry2arrow_batch.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def _batch_convert_smry2arrow(
+    ens_path: Path, ecl_base: Path, relative_output_dir: Path
+) -> None:
+    """Does batch conversion of UNSMRY files for all realizations within an ensemble."""
+    globbed_real_dirs = sorted(glob.glob(str(ens_path)))
+
+    for real_dir in globbed_real_dirs:
+        glob_expr = os.path.join(real_dir, ecl_base)
```
asnyv, line 52: Doesn't really matter, but maybe nicer to keep to `pathlib.Path` instead of mixing with `os.path` (happens a couple of places)
```suggestion
        glob_expr = Path(real_dir) / ecl_base
```

webviz_subsurface/ert_jobs/smry2arrow_batch.py
```diff
@@ -0,0 +1,103 @@
...
+        globbed_smry_files = sorted(glob.glob(glob_expr))
+        if globbed_smry_files:
+            real_output_dir = os.path.join(real_dir, relative_output_dir)
+            os.makedirs(real_output_dir, exist_ok=True)
```
asnyv, line 56: And again if staying with `pathlib.Path`. But again, no real effect 🙈
```suggestion
            real_output_dir.mkdir(parents=True, exist_ok=True)
```

anders-kiaer:
> Maybe even go straight to DeprecationWarning :sweat_smile: 

That is actually a good :bulb: (i.e. emit a `UserWarning` in this :face_with_head_bandage: temporary offer from the start)

berland:
>  "frightening"
indeed.

sigurdp:
> > Maybe even go straight to DeprecationWarning 😅
> 
> That is actually a good 💡 (i.e. emit a `UserWarning` in this 🤕 temporary offer from the start)

Added a `FutureWarning` with the following warning text:
_This script is a temporary solution for converting existing ensemble summary data for usage with webviz-subsurface, and will probably be removed in the future. New ensembles should be configured with an ert job for outputting .arrow files directly._

Maybe there should be some other wording or more specific instructions wrt how to generate the output directly?

anders-kiaer (approved):
LGTM @sigurdp once 🟢 CI 👍

PR #777, opened 2021-09-22
Added source and ensemble comparison to `VolumetricAnalysis`

tnatt:
PR which adds content to the `Source comparison` and `Ensemble comparison` tabs in `VolumetricAnalysis`.
These tabs are focused on computing differences between sources/ensembles, and is a tool for quick detection of where and why volumetric changes occur.

Both tabs have the same layout and functionality:
- Highlight data based on criteria to quickly detect where volumes changes occur
- Flexible selection of level of investigation
- Plots for visual inspection of correlation etc.
- Tables for looking at the differences for the selected response 
- Tables for looking at differences for multiple responses together to understand why volumes are changing

The` Source comparison` is useful for comparing volumes between:
- geogrid and upscaled grid from RMS. 
- static volumes and dynamic volumes 

The `Ensemble comparison` is useful for comparing volumes between:
- AHM iterations 
- Model revisions 

https://webviz-subsurface-review-tnatt.azurewebsites.net/volumetrics-analysis

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED] - YYYY-MM-DD
 ### Added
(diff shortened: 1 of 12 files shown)


HansKallekleiv (approved):
- Tested and verified working on different fields
- Code looks good 👍 

This will be a great improvement 🥇 😃

PR #709, opened 2021-07-02
Vector Calculator in Reservoir Simulation Timeseries plugin

jorgenherje:
Added usage of the vector calculator react component in reservoir simulation timeseries Dash plugin

_**Dependent on:**_
- https://github.com/equinor/webviz-subsurface-components/pull/434
- https://github.com/equinor/webviz-subsurface-testdata/pull/87

---

### Functionality

- Add new expression
- Delete expressions
- Clone expressions
- Edit expressions
- Plot expressions

PR #779, opened 2021-09-23
Allow no `predefined_expressions` in `ReservoirSimulationTimeseries`

asnyv:
Fixes a bug in the still unreleased #709 where the logic to allow missing predefined_expressions broke after rewriting the shared_settings verification steps.

### DIFF
--- a/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
+++ b/webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
@@ -316,12 +316,14 @@ folder, to avoid risk of not extracting the right data.
             line_shape_fallback
         )
 
-        # Retreive predefined expressions from configuration
-        self.predefined_expressions_path = None
-        if "predefined_expressions" in webviz_settings.shared_settings:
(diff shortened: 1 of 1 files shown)


webviz_subsurface/plugins/_reservoir_simulation_timeseries.py
```diff
@@ -316,12 +316,14 @@ def __init__(
...
-                "predefined_expressions"
-            ][predefined_expressions]
+        # Retreive predefined expressions from configuration and validate
+        self.predefined_expressions_path = (
+            None
+            if predefined_expressions is None
+            else webviz_settings.shared_settings["predefined_expressions"][
+                predefined_expressions
+            ]
+        )
```
anders-kiaer, line 326:
```suggestion
        self.predefined_expressions_path = webviz_settings.shared_settings.get("predefined_expressions", {}).get(predefined_expressions, None)
``
An alternative to consider.

anders-kiaer (approved):
LGTM 🐞 👍

PR #783, opened 2021-09-23
Added fipfile QC tab to VolumetricAnalysis

tnatt:
PR which adds a Fipfile QC tab to VolumetricAnalysis.
This tab will only appear if users inputs a fipfile.

When comparing static and dynamic volumes the fipfile is used to guide which fipnums and regions/zones to sum over to get comparables volumes. If the user makes this file manually errors can occur. This tab presents a way of QC'ing the mapping between fipnum vs zone/region and to see what data are grouped together by the fipmapper (SETs).
- this is done in both a visual manner using heatmaps and by table form 

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/plugins/_volumetric_analysis/controllers/fipfile_qc_controller.py
+++ b/webviz_subsurface/plugins/_volumetric_analysis/controllers/fipfile_qc_controller.py
@@ -1,0 +1,117 @@
+from typing import Callable
+
+import pandas as pd
+from dash import Dash, html, Input, Output
+from dash.exceptions import PreventUpdate
+import plotly.express as px
+import plotly.graph_objects as go
+import webviz_core_components as wcc
+from ..utils.table_and_figure_utils import create_table_columns, create_data_table
+
+
+def fipfile_qc_controller(
+    app: Dash,
+    get_uuid: Callable,
+    disjoint_set_df: pd.DataFrame,
+) -> None:
+    @app.callback(
+        Output(get_uuid("main-fipqc"), "children"),
+        Input(get_uuid("selections"), "data"),
+        Input(get_uuid("page-selected"), "data"),
+        Input({"id": get_uuid("main-fipqc"), "element": "display-option"}, "value"),
+    )
+    def _update_page_fipfileqc(
+        selections: dict, page_selected: str, display_option: str
+    ) -> html.Div:
+
+        if page_selected != "fipqc":
+            raise PreventUpdate
+
+        df = disjoint_set_df[["SET", "FIPNUM", "REGION", "ZONE", "REGZONE"]]
+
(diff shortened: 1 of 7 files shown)


webviz_subsurface/plugins/_volumetric_analysis/controllers/fipfile_qc_controller.py
```diff
@@ -0,0 +1,118 @@
+from typing import Callable
+import pandas as pd
+import dash
+from dash.dependencies import Input, Output
+from dash.exceptions import PreventUpdate
+import dash_html_components as html
+import plotly.express as px
+import plotly.graph_objects as go
+import webviz_core_components as wcc
+from ..utils.table_and_figure_utils import create_table_columns, create_data_table
```
anders-kiaer, line 10:
```suggestion
from typing import Callable

import pandas as pd
import dash
from dash.dependencies import Input, Output
from dash.exceptions import PreventUpdate
import dash_html_components as html
import plotly.express as px
import plotly.graph_objects as go
import webviz_core_components as wcc

from ..utils.table_and_figure_utils import create_table_columns, create_data_table
```

anders-kiaer: Also use `from dash import html` instead of `import dash_html_components as html`

webviz_subsurface/plugins/_volumetric_analysis/views/fipfile_qc_layout.py
```diff
@@ -0,0 +1,83 @@
+from typing import List
+import pandas as pd
+
+import dash_html_components as html
+import webviz_core_components as wcc
```
anders-kiaer, line 5:
```suggestion

import pandas as pd

from dash import html
import webviz_core_components as wcc
```

anders-kiaer (changes requested):

webviz_subsurface/plugins/_volumetric_analysis/views/fipfile_qc_layout.py
tnatt: Should be fixed now 👍 but I put pandas just above dash.. that is correct right?

anders-kiaer: Looks good :) The convention is std.lib before other packages before itnernal in this package :)

https://www.python.org/dev/peps/pep-0008/#imports

> Imports should be grouped in the following order:
> 
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.
> You should put a blank line between each group of imports.

anders-kiaer (approved):

PR #784, opened 2021-09-23
Layout change Ensemble comparison tab `VolumetricAnalysis`

tnatt:
For ensembles it doesn't make sense to compare on realization level like it does for sources. 

This PR removes the realization vs diff plot from the layout and removes the option of grouping on realization level for the `Ensemble comparison` tab in `VolumetricAnalysis`.

Edit to the not released #777

| Source comparison layout | Ensemble comparison layout |
| -----------------------------|---------------------------------|
|image | image |

### DIFF
--- a/webviz_subsurface/plugins/_volumetric_analysis/views/comparison_layout.py
+++ b/webviz_subsurface/plugins/_volumetric_analysis/views/comparison_layout.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from dash import html, dcc, dash_table
 import plotly.graph_objects as go
 import webviz_core_components as wcc
@@ -38,11 +39,12 @@ def comparison_main_layout(uuid: str) -> html.Div:
 
 
 def comparison_qc_plots_layout(
-    fig_diff_vs_real: go.Figure,
+    fig_diff_vs_real: Optional[go.Figure],
     fig_corr: go.Figure,
     fig_diff_vs_response: go.Figure,
     barfig: go.Figure,
 ) -> html.Div:
+    real_plot = fig_diff_vs_real is not None
     return html.Div(
         children=[
             html.Div(
@@ -51,21 +53,23 @@ def comparison_qc_plots_layout(
                     style={"height": "22vh"},
                     figure=fig_diff_vs_real,
                 )
+                if real_plot
+                else [],
             ),
             wcc.FlexBox(
-                style={"height": "32vh"},
+                style={"height": "32vh" if real_plot else "54vh"},
                 children=[
                     wcc.FlexColumn(
                         children=wcc.Graph(
                             config={"displayModeBar": False},
-                            style={"height": "31vh"},
+                            style={"height": "31vh" if real_plot else "53vh"},
                             figure=fig_diff_vs_response,
                         )
                     ),
                     wcc.FlexColumn(
                         children=wcc.Graph(
                             config={"displayModeBar": False},
-                            style={"height": "31vh"},
+                            style={"height": "31vh" if real_plot else "53vh"},
                             figure=fig_corr,
                         )
                     ),
@@ -152,6 +156,14 @@ def comparison_selections(
                         selector_label="value2",
                         value=elements[1] if compare_on == "SOURCE" else elements[-1],
                         elements=elements,
+                    ),
+                    html.Div(
+                        f"Difference = {compare_on.capitalize()} B - {compare_on.capitalize()} A",
+                        style={
+                            "font-size": "15px",
+                            "margin-top": "5px",
+                            "color": "#007079",
+                        },
                     ),
                     response_selector(volumemodel, uuid, tab),
                     group_by_selector(volumemodel, uuid, tab),
--- a/webviz_subsurface/plugins/_volumetric_analysis/controllers/comparison_controllers.py
+++ b/webviz_subsurface/plugins/_volumetric_analysis/controllers/comparison_controllers.py
@@ -131,14 +131,15 @@ def comparison_callback(
             filter_info="SOURCE" if compare_on != "SOURCE" else "ENSEMBLE",
         )
 
-    diffdf_real = create_comparison_df(
-        volumemodel,
-        compare_on=compare_on,
-        selections=selections,
-        responses=[selections["Response"]],
-        groups=groupby + (["REAL"] if "REAL" not in groupby else []),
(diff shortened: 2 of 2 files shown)


asnyv:
A "stupid" question: If we are not comparing by realization, what are we then comparing by? Mean values?

I haven't followed the latest updates to this plugin, so I'm not completely up-to-date (and might very well say something wrong here 😅): At least I see now in the demo that it is possible to make comparisons without REAL (if you have not selected it), and that the volumes over the realizations are then collapsed into a single value. I assume they are mean values over the ensemble, but think it should be specified 🤔 Ideally also with the option of changing aggregation between e.g. mean, various percentiles, max/min, stddev and so on

tnatt:
> A "stupid" question: If we are not comparing by realization, what are we then comparing by? Mean values?
> 
> I haven't followed the latest updates to this plugin, so I'm not completely up-to-date (and might very well say something wrong here 😅): At least I see now in the demo that it is possible to make comparisons without REAL (if you have not selected it), and that the volumes over the realizations are then collapsed into a single value. I assume they are mean values over the ensemble, but think it should be specified 🤔 Ideally also with the option of changing aggregation between e.g. mean, various percentiles, max/min, stddev and so on
> 
> Also: for the concept of "delta ensembles" in other plugins, we do allow differences per realization for "anything" (including inplace volumes), and I think it could be useful here too. E.g. if you have run two similar ensembles, where you have the same seed per realization and only vary a contact or something like that, and want to compare the outcomes down to the realization level.

Yes, you are correct it is the mean values only at the moment (could probably specify that somewhere 😃). And being able to switch between  aggregation to compare on when it comes to ensembles makes sense and will be a nice further implementation 👍. 
But when it comes to investigation on realization level, in my mind it is not often you want to do that for ensembles.. for ahm it doesn't make sense, and if you do any structural changes to your static model the same seed number will give you different results. 
But how about I keep REAL as an option in the "level of investigation", then the user at least have the option of comparing per REAL in the tables, but I still drop the plot? Because the plot will in most cases be misleading (a lot of highlighted realization without any value).

tnatt:
> > A "stupid" question: If we are not comparing by realization, what are we then comparing by? Mean values?
> > I haven't followed the latest updates to this plugin, so I'm not completely up-to-date (and might very well say something wrong here 😅): At least I see now in the demo that it is possible to make comparisons without REAL (if you have not selected it), and that the volumes over the realizations are then collapsed into a single value. I assume they are mean values over the ensemble, but think it should be specified 🤔 Ideally also with the option of changing aggregation between e.g. mean, various percentiles, max/min, stddev and so on
> > Also: for the concept of "delta ensembles" in other plugins, we do allow differences per realization for "anything" (including inplace volumes), and I think it could be useful here too. E.g. if you have run two similar ensembles, where you have the same seed per realization and only vary a contact or something like that, and want to compare the outcomes down to the realization level.
> 
> Yes, you are correct it is the mean values only at the moment (could probably specify that somewhere 😃). And being able to switch between aggregation to compare on when it comes to ensembles makes sense and will be a nice further implementation 👍.
> But when it comes to investigation on realization level, in my mind it is not often you want to do that for ensembles.. for ahm it doesn't make sense, and if you do any structural changes to your static model the same seed number will give you different results.
> But how about I keep REAL as an option in the "level of investigation", then the user at least have the option of comparing per REAL in the tables, but I still drop the plot? Because the plot will in most cases be misleading (a lot of highlighted realization without any value).

Did not mean to close this 😄

asnyv:
> Did not mean to close this 😄

Haha yeah I was wondering for a second if you were that trigger happy 😉

>But when it comes to investigation on realization level, in my mind it is not often you want to do that for ensembles.. for ahm it doesn't make sense, and if you do any structural changes to your static model the same seed number will give you different results.
But how about I keep REAL as an option in the "level of investigation", then the user at least have the option of comparing per REAL in the tables, but I still drop the plot? Because the plot will in most cases be misleading (a lot of highlighted realization without any value).

I agree that for AHM it doesn't make sense, assuming you have uncertainty in the static model as part of the workflow (more than just multipliers). But for static models without update of structure and parameter fields (which I believe is quite common for older models?), the static model could be constant between the iterations? Only updating e.g. fluid contact and some multipliers on pore volume, properties and so on? In which case I would think this comparison is relevant also between ensembles
But I can think I agree that the topmost plot could be removed in that case. Think the other plots should cover the need 😊 

Also when I remember it (both for sources and ensembles): could you specify which way the diff goes? If e.g. the labels say geogrid - simgrid instead of diff, or that it somewhere on the dashboard is clearly stated?

tnatt:
> I agree that for AHM it doesn't make sense, assuming you have uncertainty in the static model as part of the workflow (more than just multipliers). But for static models without update of structure and parameter fields (which I believe is quite common for older models?), the static model could be constant between the iterations? Only updating e.g. fluid contact and some multipliers on pore volume, properties and so on? In which case I would think this comparison is relevant also between ensembles
> But I can think I agree that the topmost plot could be removed in that case. Think the other plots should cover the need 😊

Ok, I think we agree then that the plot is dropped but we'll keep the option of investigating on REAL level 🙂 That should cover most needs. I'll do the change!

> Also, while I remember it (both for sources and ensembles): could you specify which way the diff goes? If e.g. the labels say geogrid - simgrid instead of diff, or that it somewhere on the dashboard is clearly stated?

It is stated in the two tables which way the diff has been calculated, but I can include it as a text in the selections to make it more visible 👍 that is a good idea!

asnyv:
> It is stated in the two tables which way the diff has been calculated, but I can include it as a text in the selections to make it more visible 👍 that is a good idea!

Aah ok, I looked at plots and not the tables 😉 but that looks good 😊 think it is important that the definition of the difference is visible no matter which view option you've chosen (plot, tables etc), so having it explicit in the "menu" on the left is a good idea 👍

webviz_subsurface/plugins/_volumetric_analysis/views/comparison_layout.py
```diff
@@ -1,3 +1,4 @@
+from typing import Optional
 from dash import html, dcc, dash_table
```
anders-kiaer, line 2:
```suggestion
from typing import Optional

from dash import html, dcc, dash_table
```

anders-kiaer (approved):
One minor comment otherwise LGTM! 👍

PR #788, opened 2021-09-24
Prevent mixing volumes from different sensitivities in `VolumetricAnalysis`

tnatt:
User request.
Handle SENSNAME in same manner as SOURCE/ENSEMBLE by not allowing more than one SENSNAME to be selected as filter if the user has not grouped on this column. This is to avoid misleading statistical numbers.

The standard name for the SENSNAME with "base case" simulations is "rms_seed", this is used as default if present when SENSNAME has not been grouped on.


anders-kiaer (approved):
Add changelog entry, or is this also a change on something not yet released?

tnatt:
> Add changelog entry, or is this also a change on something not yet released?

Well spotted, I forgot 👍 will do!

PR #791, opened 2021-09-24
Ensure map bounds are updated when switching attribute in `SurfaceViewerFMU`

HansKallekleiv:
When switching between attributes with different geometry the map bounds are not updated. There seems to be an issue using `updateMode="update"`. 
Switching to `updateMode="replace"` fixes the issue.

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


anders-kiaer (approved):
Add changelog entry 😄 , otherwise LGTM.

PR #792, opened 2021-09-27
Update tests based on change in testdata set

anders-kiaer:
Related to https://github.com/equinor/webviz-subsurface-testdata/pull/105.


asnyv (approved):
✔

PR #794, opened 2021-09-27
Bugfix sensitivity run detection `ParametersModel` 

tnatt:
Bugfix in the sensitivity run detection logic in `ParametersModel`. 
Desired behavior, if only ensembles with single montecarlo sensitivities the run should not be flagged as a sensitivity run.


anders-kiaer (approved):
Add changelog entry (this file was changed in August last time it looks like?) - otherwise LGTM 👍🐞🚢

Issue #800, opened 2021-09-28
`VolumetricAnalysis` - remove totals volume if present in the input


tnatt:
If `Totals` is present as an element in an selector users have to filter out this element to prevent volumes being doubled. 
It would be good to remove "Totals" from all selectors if present.

e.g. 

| ZONE | STOIIP_OIL |
|--------|-------------|
| topzone | 10000 |
| basezone | 20000 |
| Totals | 30000 |

tnatt:
Closed in #802

PR #802, opened 2021-09-29
Bugfixes `VolumetricAnalysis`

tnatt:
PR with small bugfixes for `VolumetricAnalysis`:

Comparison pages:
1. Handle None in the Input components by setting a default value of 0
2. Fix issue to avoid data from the different FLUID_ZONEs stacking upon eachother in the barfigure
3. Use "Total" as label in the barfigure if no "level of investigation" has been selected

Other:
1. Removed "BO" and "BG" as option in tornado as they will cause error if multiple FLUID_ZONEs exists
2. Print out "valid static columns" for the user if the volume type most likely is static but contains either custom columns or not standard column names. 
3. Remove TOTAL columns if both static and dynamic volumes have been given as input to avoid misleading comparison (they often represent different things)
4. Remove rows with total volumes from selectors if they are present to avoid doubling of volumes #800

Most issues are bugfixes on non-released code, except removing "BO" and "BG" from the tornados . I'll add a changelog for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#761](https://github.com/equinor/webviz-subsurface/pull/761) - Store `xtgeo.RegularSurface` as bytestream instead of serializing to `json`.
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


anders-kiaer (approved):
👍🐛🙂

PR #830, opened 2021-10-29
Fixed doc and polygon selector

rnyb:
Fixed bug in doc + extended doc to explain polygon input.

For the polygon, the multi select option is now set to False as it
caused issues when combining multiple polygon-files. Also set initial
polygon value to empty for slightly better performance at start-up.

Also added minor edit to CHANGELOG.md


HansKallekleiv (changes requested):
Some lines that should be removed, otherwise LGTM!

HansKallekleiv (approved):

PR #833, opened 2021-11-01
Bugfixes to `VolumetricAnalysis`

tnatt:
PR with bugfixes to the `VolumetricAnalysis` plugin:
- Handle empty dataframes after filtering in the `comparison` pages 
- Handle empty dataframes and missing tornado reference after filtering in the `tornadoplots` page 
- Fix error in the "update" logic of pages – set the initial_load state to False after a user selection is made
- Ensure the filter for the selected subplot column is set to multi in the `tornadoplots` page


HansKallekleiv (approved):
LGTM 💡

PR #838, opened 2021-11-02
Fix timing issue in `AssistedHistoryMatchingAnalysis`

### LINKED ISSUE #837: `AssistedHistoryMatchingAnalysis` crashes on initialization due to missing object
Caused by the `clickData` property in a callback being the output of another callback, which is not guaranteed to be finished when the callback is fired the first time.
https://github.com/equinor/webviz-subsurface/blob/9df1a7f0c9345795b5545b5796bf8f0347fe08cb/webviz_subsurface/plugins/_assisted_history_matching_analysis.py#L261

Most likely always been a bug, but could be that it has not been causing issues for users before the new menu was implemented.

asnyv:
### Contributor checklist

- [X] :tada: This PR closes #837.


HansKallekleiv (approved):

anders-kiaer (approved):
👍

PR #841, opened 2021-11-04
Bugfixes and improved hoverlabels for `Tornado component`

### LINKED ISSUE #839: Bugs in tornado for sensitivities where both cases are on the same side of the reference value
**How to reproduce**
The behavior is observable in the [demo for ReservoirSimulationTimeseriesOneByOne](https://webviz-subsurface-example.azurewebsites.net/time-series-analysis) for the multregt_mc sensitivity

For the multregt_mc sensitivity, both the high and low of the outputs end up lower than the reference average (using RMS_SEED as reference).

When using "Relative value" as scale, only the low case is shown for the labels:
image
Also, when hovering the low value is in fact the difference between the high and the low of the sensitivity, and not relative to the reference. And the high is simply 0.
image
When switching to "True value" it gets a bit worse. The bar is now starting at the reference value, and the length is equal to the difference between the high and the low (so it is neither the low relative to the ref, nor the high)
image
This is also shown when hovering, where the length of the bar is used as low, and high again is 0.
image

**Extra**
It is also a bit confusing with the terms low/high in the hover labels and , as they are also often used as cases in a sensitivity, but they are not necessarily corresponding (as the term low/high is generally referring to low/high values for the parameter, while that might have an inverse effect on the response we are looking at). Should consider to use the case names instead of low/high.

Have also gotten a user comment that the term "True value" is not ideal, and that we should consider to go back to e.g. "absolute" or maybe "actual" as the antonym  to "relative", as true/truth can be associated with the actual truth, and not modelled results.

tnatt:
PR with bugfixes 

improved hoverlabel for tornados and bugfixes
- show correct x value
- use senscase instead of low/high
- only show one senscase for single case sensitivities
- fix axis range bug
- fix issue in tornado when True value was selected and both high and low on one side of reference.

one bugfix in `VolumetricAnalysis`
- fix error when showing ensembles as subplot and some ensembles do not have the reference sensitivity

---

### Contributor checklist

- [ ] :tada: This PR closes #839.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#809](https://github.com/equinor/webviz-subsurface/pull/809) - `GroupTree` - added more statistical options (P10, P90, P50/Median, Max, Min). Some improvements to the menu layout and behaviour
 
 ### Fixed
+- [#841](https://github.com/equinor/webviz-subsurface/pull/841) - Bugfixes and improved hoverlabels for `Tornado component`.
 - [#833](https://github.com/equinor/webviz-subsurface/pull/833) - Fixed errors in `VolumetricAnalysis` related to empty/insufficient data after filtering in the `tornadoplots` and `comparison` tabs.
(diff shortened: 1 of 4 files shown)


asnyv (approved):
👍

PR #842, opened 2021-11-07
Groups as leaf nodes in the `GroupTree` plugin

### LINKED ISSUE #787: GroupTree plugin: support groups as leaf nodes
We need to support cases without wells. Thinking especially for coupled models where the well data is stored in other sets of summary data (per "slave"), and where this visualization might be especially valuable do to complex network structures. I don't think supporting gathering data from slave models is needed (and for sure not in the first iteration), but it would be nice to be allowed to not have wells in the model (so that the lowermost level is a set of groups). I see that a decent chunk of the logic is based on wells as the leaves of the network, so it might be ok to let it be a requirement for the first iteration if it is very cumbersome, but it is for sure something we have to support. You may also have leaves which are not wells outside network models, e.g. if you are using GSATPROD or GCONSUMP.

lindjoha:
Fixes #787. 

The plugin will now identify group leaf nodes and classify them as producing/injecting/other by using the summary data. If f.ex the production summary data is not exported it will not be recognized as a producer even if it is.


tests/unit_tests/plugin_tests/test_grouptree.py
```diff
@@ -0,0 +1,125 @@
...
+        "CHILD",
+        "KEYWORD",
+        "PARENT",
+        "IS_PROD",
+        "IS_INJ",
+        "IS_OTHER",
+    ]
+    output = add_nodetype_for_ens(gruptree, smry)
+    print("output")
+    print(output)
```
anders-kiaer, line 123: Remove?

webviz_subsurface/plugins/_group_tree/group_tree_data.py
```diff
@@ -395,42 +395,52 @@ def get_nodedict(gruptree: pd.DataFrame, nodename: str) -> Dict[str, Any]:
...
     nodes["IS_OTHER"] = nodes["CHILD"].map(is_other_map)
 
-    # Recursively find well types of all wells connected to the group node
+    # Recursively find well types of all leaf nodes connected to the group node
     # Deduce group node type from well types
-    groupnodes = nodes[nodes["KEYWORD"] != "WELSPECS"]
-    for _, groupnode in groupnodes.iterrows():
-        wells_are_prod, wells_are_inj, wells_are_other = get_leaf_well_types(
-            groupnode["CHILD"], nodes
+    nonleafs = nodes[nodes["IS_LEAF"].isin([False])]
```
anders-kiaer, line 436: There are perhaps shorter notation for this? Does the "numpy" syntax also work for pandas? I.e.
```suggestion
    nonleafs = nodes[~nodes["IS_LEAF"]]
```

anders-kiaer (changes requested):
Nice solution. A couple of minor comments only. Add a changelog entry?

webviz_subsurface/plugins/_group_tree/group_tree_data.py
lindjoha: Yes it works! I was searching for something like that, but I couldn't find it :)

anders-kiaer (approved):

PR #845, opened 2021-11-15
Added realplot to tornadopage in VolumetricAnalysis

tnatt:
Improvements to the tornado tab in `VolumetricAnalysis`

- fixed and improved formatting of numbers not suitable for si format - both in figure and in table
- added option of turning on/off reference on tornadofigure
- added possibility of visualizing a realization plot colored by sensname below the tornado figure

image


HansKallekleiv (approved):
👍

PR #847, opened 2021-11-17
Refactoring `WellCompletions` code

lindjoha:
The `WellCompletions `plugin code is refactored according to the best practice plugin in `webviz-plugin-boilerplate`. The code is split into four files: `_business_logic.py`, `_callbacks.py`, `_layout.py`, `_plugin.py`.

Clientside store functionality is added so that it will remember which ensemble is selected in the dropdown.


anders-kiaer (approved):
VERY nice @lindjoha! 🏗️

PR #851, opened 2021-11-22
Added combined sensname-senscase column to `VolumetricAnalysis`

tnatt:
PR to create a new column `SENSNAME_CASE` which is a combination of the `SENSNAME` and `SENSCASE`
Benefit with a combo column:
- Easier filtering when only one column 
  - no invalid selection possible  
- Easier to create plots with plotly express 
  - plotting SENSNAME and coloring on SENSCASE makes very thin bars/boxes...

Also:
- Fixed issue that made it impossible to create barplots per REAL


anders-kiaer (approved):
LGTM. 👍

PR #855, opened 2021-11-24
Allow mixing sensitivity and gen_kw ensembles in `ParametersModel`

tnatt:
PR to support a mix of sensitivity ensembles and gen_kw ensembles in `ParametersModel`. 
A dice is added as SENSNAME to gen_kw ensembles for easy distinction from sensitivity ensembles:
image

### DIFF
--- a/webviz_subsurface/_models/parameter_model.py
+++ b/webviz_subsurface/_models/parameter_model.py
@@ -106,10 +106,11 @@ class ParametersModel:
         if "SENSNAME" not in self._dataframe:
             return False
 
-        if self.dataframe["SENSNAME"].isnull().values.any():
-            raise ValueError(
-                "Ensembles with and without sensitivity data mixed - this is not supported!"
-            )
+        # if mix of gen_kw and sensitivity ensembles add
+        # dummy sensitivvity columns to gen_kw ensembles
+        gen_kw_mask = self._dataframe["SENSNAME"].isnull()
+        self._dataframe.loc[gen_kw_mask, "SENSNAME"] = "🎲"
+        self._dataframe.loc[gen_kw_mask, "SENSCASE"] = "p10_p90"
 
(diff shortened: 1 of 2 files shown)


webviz_subsurface/_models/parameter_model.py
```diff
@@ -106,11 +107,17 @@ def _check_if_sensitivity_run(self) -> bool:
         if "SENSNAME" not in self._dataframe:
             return False
 
-        if self.dataframe["SENSNAME"].isnull().values.any():
+        gen_kw_mask = self._dataframe["SENSNAME"].isnull()
+        if gen_kw_mask.values.any() and not allow_ensemble_type_mix:
             raise ValueError(
                 "Ensembles with and without sensitivity data mixed - this is not supported!"
             )
```
anders-kiaer, line 114: Should this error message be updated slightly now? E.g. replace "this is not supported" with "this is not enabled".

anders-kiaer (commented):
LGTM. 👍 One comment below only from my point of view.

For my understanding: Are there use cases where it is valuable for the user _not_ to allow mixing? If not, should it always be allowed and we can remove `allow_ensemble_type_mix` input argument?

tnatt:
> LGTM. 👍 One comment below only from my point of view.
> 
> For my understanding: Are there use cases where it is valuable for the user _not_ to allow mixing? If not, should it always be allowed and we can remove `allow_ensemble_type_mix` input argument?

Originally I just removed the Error, but than I thought we might have plugins were it might be important to only have one.. but thinking about it again, as long as we provide sensitivity columns to non-sensitivity ensembles it should not be a problem to include them in any plugin :) I'll clean up the PR and remove that option.. can rather put it back in if ever needed 👍

anders-kiaer (approved):
🎄

PR #863, opened 2021-11-29
Second bugfix on sensitivity comparison in VolumetricAnalysis

tnatt:
Was too quick with the bugfix on friday.. 🤔 
..need to ensure "SENSNAME_CASE" is not present as index, otherwise not possible to compare sensitivities across ensembles


anders-kiaer (approved):
👍

PR #873, opened 2021-12-07
 `GroupTree` plugin: bugfix related to BRANPROP group leaf nodes

lindjoha:
This bug relates to node type classification of BRANPROP group leaf nodes. Group leaf nodes are classified by using summary data, but BRANPROP nodes does not have injection summary vectors by definition.  Therefore the `create_leafnodetype_maps` function was crashing.

This bug is fixed by setting suminj = 0 for BRANPROP group `nodes.`


anders-kiaer (approved):
👍

PR #878, opened 2021-12-09
Use regular numpy for all map statistic calculations

HansKallekleiv:
Regression fix for #875 
Use regular numpy calculations for all map statistics calculation, not just percentiles.


anders-kiaer (approved):
:+1:

PR #888, opened 2021-12-15
Seismic misfit improvements

rnyb:
Improved performance when using polygon files with a large number of polygons.
Also changed default value of sim_path.

### DIFF
--- a/webviz_subsurface/plugins/_seismic_misfit.py
+++ b/webviz_subsurface/plugins/_seismic_misfit.py
@@ -4,7 +4,6 @@ import glob
 import logging
 import math
 import re
-import time
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -162,7 +161,7 @@ class SeismicMisfit(WebvizPluginABC):
         webviz_settings: WebvizSettings,
         ensembles: List[str],
         attributes: List[str],
-        attribute_sim_path: str = "sim2seis/output/4d_attribute_maps/",
+        attribute_sim_path: str = "share/results/maps/",
         attribute_obs_path: str = "../../share/observations/seismic/",
         obs_mult: float = 1.0,
         sim_mult: float = 1.0,
@@ -2057,7 +2056,6 @@ def update_misfit_plot(
 
     # max_diff = find_max_diff(df)
     max_diff = None
-    min_diff = None
     figures = []
 
     for ens_name, ensdf in df.groupby("ENSEMBLE"):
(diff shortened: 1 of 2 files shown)


webviz_subsurface/plugins/_seismic_misfit.py
```diff
@@ -161,7 +161,7 @@ def __init__(
         webviz_settings: WebvizSettings,
         ensembles: List[str],
         attributes: List[str],
-        attribute_sim_path: str = "sim2seis/output/4d_attribute_maps/",
+        attribute_sim_path: str = "share/results/maps/",
```
anders-kiaer, line 164: Is this a new default output path? Do we need to do any changes in `testdata` or is it ok as is?

webviz_subsurface/plugins/_seismic_misfit.py
```diff
@@ -2109,9 +2109,9 @@ def update_misfit_plot(
             x="REAL",
             y="ABSDIFF",
             title=ens_name,
-            range_y=[min_diff * 0.30, max_diff * 1.05],
+            range_y=[min_diff * 0.0, max_diff * 1.05],
             color="ABSDIFF",
-            range_color=[min_diff * 0.30, max_diff * 1.00],
+            range_color=[min_diff * 0.0, max_diff * 1.00],
```
anders-kiaer, line 2114: What is the reason for this change? Should it just be `0.0` instead of `min_diff * 0.0` which is anyway 0?

webviz_subsurface/plugins/_seismic_misfit.py
```diff
@@ -2257,21 +2257,20 @@ def update_obsdata_map(
...
-            fig.add_trace(
+        for poly_id, polydf in df_polygon.groupby("POLY_ID"):
+            fig.append_trace(
                 go.Scattergl(
                     x=polydf["X_UTME"],
                     y=polydf["Y_UTMN"],
                     mode="lines",
                     line_color="RoyalBlue",
-                    name=poly_id,
+                    name="pol" + str(poly_id),
```
anders-kiaer, line 2267:
```suggestion
                    name=f"pol{poly_id}",
```

webviz_subsurface/plugins/_seismic_misfit.py
```diff
@@ -2515,21 +2514,20 @@ def update_obs_sim_map_plot(
...
```
anders-kiaer, line 2524:
```suggestion
                    name=f"pol{poly_id}",
```

anders-kiaer (changes requested):
LGTM @rnyb. Some small comments below. Also add a changelog entry?

webviz_subsurface/plugins/_seismic_misfit.py
rnyb: Yes, this is a new default output. I recently made a change in testdata webviz config file to prepare for this change so it should work as is. However, we could consider also moving the data itself from `sim2seis/output/4d_attribute_maps` to `share/results/maps/`, but it is not necessary for this to work.

webviz_subsurface/plugins/_seismic_misfit.py
rnyb: I changed this to 0 as i ran into a case where some realization results were initially hidden, but i agree this was a sloppy change. I will drop the min_diff.

anders-kiaer (approved):

PR #895, opened 2021-12-20
General improvements to VolumetricAnalysis

tnatt:
Several improvements to VolumetricAnalysis, some based on of user-feedback.

**Main changes `Inplace distributions` tab**

- Merged functionality in pages `1 plot/1 table` and `Custom plotting`, in order to make it less confusing for the users what page to use, and also simplify the code logic. The only real difference between the two pages was that the user could make subplots in the `Custom plotting`. This is still available, and an option to hide the table below is added to make more space for the subplots. 
- Added useful "colorby" feature to the barplots in `Plots per zone/region` page. If colorby is selected the piechart is hidden.
- Change to returning entire layout for the pages, instead of individual figures --> if a selection/filter changes all plots needs to be updated anyway. This helps improving the readability of the code and reduces a lot of code lines.
- Remove "REAL" as colorby option for plot types other than "scatter"
- Remove the fluid annotation automatically if the x-response is a HC-response
- Added shading to `Convergence plot`s to indicate missing realizations, and added more explanatory hoverinfo.  

**Other general changes**

- Adjust heights of different views to remove overflow
- `Comparions tab`: extend x-axis range by 10% to fix formatting issue when only one data point, and add dotted lines indicating the "highlight threshold" for all plots with difference in percent on the y-axis
- `Tornado tab`: removed confusing "casetype" in hover for the "realization plot" 
- Only calculate "NTG" and "PORO_NET" columns if "NET" varies from "BULK"

 

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [#889](https://github.com/equinor/webviz-subsurface/pull/889) - Added `rel_file_pattern` argument to .arrow related factory methods in EnsembleSummaryProviderFactory.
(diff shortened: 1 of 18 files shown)


tnatt:
Will add changelog tomorrow 🙂

HansKallekleiv (changes requested):
LGTM :)
Approved after changelog is added.

HansKallekleiv (approved):

PR #897, opened 2021-12-21
Add Python 3.10 to CI

anders-kiaer:
...currently only in order to smoke out which dependencies are missing wheels.


anders-kiaer:
Blocked by https://github.com/equinor/xtgeo/issues/741.

anders-kiaer:
Blocked by https://github.com/OPM/opm-common/issues/2969.

PR #898, opened 2021-12-21
Format dates in download for resampled date in SimulationTimeSeries

asnyv:
The old `ReservoirSimulationTimeSeries` has some logic to modify the `DATE` column when downloading the special vectors that we calculate from cumulatives over a certain interval in time (those starting with AVG_ and INTVL_). The logic was introduced to reduce risk of misunderstandings from users related to whether the date was e.g. at the beginning of the interval, at the end, or the middle. Instead, the `DATE` axis was reduced down to what describes the interval best. A yearly interval would leave only `YYYY`, a monthly would give `YYYY-MM` and daily would give `YYYY-MM-DD`.

This PR implements the same logic for the new `SimulationTimeSeries` plugin, though now using the `Frequency` dataclass + extended to support weeks and quarters.

### DIFF
--- a/webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
+++ b/webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
@@ -151,7 +151,7 @@ def rename_vector_from_cumulative(vector: str, as_rate: bool) -> str:
 # pylint: disable=too-many-return-statements
 def datetime_to_intervalstr(date: datetime.datetime, freq: Frequency) -> Optional[str]:
     if date is None:
-        return date
+        return None
 
     if freq == Frequency.DAILY:
         return f"{date.year}-{date.month:02d}-{date.day:02d}"
--- a/webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
+++ b/webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
@@ -1,5 +1,10 @@
+import datetime
+from typing import Optional
+
 import numpy as np
 import pandas as pd
+
+from webviz_subsurface._providers import Frequency
 
 ###################################################################################
 # NOTE: This code is a copy and modification of
@@ -141,3 +146,25 @@ def rename_vector_from_cumulative(vector: str, as_rate: bool) -> str:
         if as_rate
         else f"INTVL_{vector}"
     )
+
+
+# pylint: disable=too-many-return-statements
+def datetime_to_intervalstr(date: datetime.datetime, freq: Frequency) -> Optional[str]:
+    if date is None:
+        return date
+
+    if freq == Frequency.DAILY:
+        return f"{date.year}-{date.month:02d}-{date.day:02d}"
+    if freq == Frequency.WEEKLY:
+        # Note: weekyear may differ from actual year for week numbers 1,52,53.
+        (weekyear, week, _) = date.isocalendar()
+        return f"{weekyear}-W{week:02d}"
+    if freq == Frequency.MONTHLY:
+        return f"{date.year}-{date.month:02d}"
+    if freq == Frequency.QUARTERLY:
+        return f"{date.year}-Q{(date.month//3+1)}"
+    if freq == Frequency.YEARLY:
+        return f"{date.year}"
+
+    # Using isoformat if frequency is none of the listed instead of error.
+    return date.isoformat()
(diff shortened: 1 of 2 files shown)


webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
```diff
@@ -141,3 +146,25 @@ def rename_vector_from_cumulative(vector: str, as_rate: bool) -> str:
         if as_rate
         else f"INTVL_{vector}"
     )
+
+
+# pylint: disable=too-many-return-statements
+def datetime_to_intervalstr(date: datetime.datetime, freq: Frequency) -> Optional[str]:
+    if date is None:
+        return date
```
anders-kiaer, line 154:
```suggestion
    if date is None:
        return None
```
Maybe slightly more readable?

webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
```diff
@@ -141,3 +146,25 @@ def rename_vector_from_cumulative(vector: str, as_rate: bool) -> str:
...
+    if freq == Frequency.DAILY:
+        return f"{date.year}-{date.month:02d}-{date.day:02d}"
+    if freq == Frequency.WEEKLY:
+        # Note: weekyear may differ from actual year for week numbers 1,52,53.
+        (weekyear, week, _) = date.isocalendar()
+        return f"{weekyear}-W{week:02d}"
+    if freq == Frequency.MONTHLY:
+        return f"{date.year}-{date.month:02d}"
+    if freq == Frequency.QUARTERLY:
+        return f"{date.year}-Q{(date.month//3+1)}"
```
anders-kiaer, line 165:
```suggestion
        return f"{date.year}-Q{1 + (date.month-1)//3}"
```

webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
```diff
@@ -141,3 +146,25 @@ def rename_vector_from_cumulative(vector: str, as_rate: bool) -> str:
...
     )
+
+
+# pylint: disable=too-many-return-statements
+def datetime_to_intervalstr(date: datetime.datetime, freq: Frequency) -> Optional[str]:
+    if date is None:
+        return date
+
+    if freq == Frequency.DAILY:
+        return f"{date.year}-{date.month:02d}-{date.day:02d}"
```
anders-kiaer, line 157:
```suggestion
        return date.strftime("%Y-%m-%d")
```

webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
```diff
@@ -141,3 +146,25 @@ def rename_vector_from_cumulative(vector: str, as_rate: bool) -> str:
...
+    if freq == Frequency.WEEKLY:
+        # Note: weekyear may differ from actual year for week numbers 1,52,53.
+        (weekyear, week, _) = date.isocalendar()
+        return f"{weekyear}-W{week:02d}"
+    if freq == Frequency.MONTHLY:
+        return f"{date.year}-{date.month:02d}"
```
anders-kiaer, line 163:
```suggestion
        return date.strftime("%Y-%m")
```

anders-kiaer (commented):
A couple of comments to consider :date:, otherwise LGTM :+1:

webviz_subsurface/plugins/_simulation_time_series/utils/from_timeseries_cumulatives.py
asnyv: Good catch! In my use case I had already ensured that we only use the first month of the quarter, so the `-1` wasn't really necessary, but I agree that it is more robust for later use cases :)

asnyv:
General comment when it comes to string formatting of `datetime`'s @anders-kiaer: I considered `strftime`, but found it a bit easier to follow the logic between the options when they all had the same structure. But considering it is only quarters that are not supported directly in `strftime`, it is of course an option to have that one as the odd one out. 
Just felt it was more readable to write it out like this, but open to change it 😉

anders-kiaer:
Sounds good, no strong opinions from my side as well between `strftime` and a more explicit (but also longer) `f`-string. I might be opinionated by having used `strftime` a lot before. 🙂 Approved 👍

anders-kiaer (approved):

PR #913, opened 2022-01-06
`SimulationTimeSeries`: Add statistics after realizations

asnyv:
Adds the statistics after the individual realizations rather than before, meaning that the statistics will be placed on top (though only for the given vector/ensemble. Additional vectors/ensembles will still be placed on top). Makes it easier to see statistics when plotting both (note difference in the area to the lower left with very high density of realizations covering all of the statistics).
Example:
Old behavior:
image

New behavior:
image

### DIFF
--- a/webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
+++ b/webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
@@ -319,16 +319,7 @@ def plugin_callbacks(
                         fanchart_options,
                     )
                 if visualization == VisualizationOptions.STATISTICS_AND_REALIZATIONS:
-                    # Configure line width and color scaling to easier separate
-                    # statistics traces and realization traces
-                    vectors_statistics_df = create_vectors_statistics_df(vectors_df)
-                    figure_builder.add_statistics_traces(
-                        vectors_statistics_df,
-                        ensemble,
-                        statistics_options,
-                        line_width=3,
-                    )
-                    # Show selected realizations on top - only filter df if realizations filter
+                    # Show selected realizations - only filter df if realizations filter
                     # query is not performed
                     figure_builder.add_realizations_traces(
                         vectors_df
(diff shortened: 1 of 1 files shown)


webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
```diff
@@ -337,6 +328,15 @@ def _update_graph(
                         ensemble,
                         color_lightness_scale=150.0,
                     )
+                    # Configure line width and color scaling + add statistics on top
+                    # to easier separate statistics traces and realization traces
```
jorgenherje, line 332: Correct comment to `# Add statistics on top`

webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
```diff
@@ -319,16 +319,7 @@ def _update_graph(
...
-                    vectors_statistics_df = create_vectors_statistics_df(vectors_df)
-                    figure_builder.add_statistics_traces(
-                        vectors_statistics_df,
-                        ensemble,
-                        statistics_options,
-                        line_width=3,
-                    )
-                    # Show selected realizations on top - only filter df if realizations filter
+                    # Show selected realizations - only filter df if realizations filter
                     # query is not performed
```
jorgenherje, line 323: Correct comment 

```
# Configure line width and color scaling to easier separate
# statistics traces and realization traces
# Show selected realizations - only filter df if realizations filter
# query is not performed
```

jorgenherje (changes requested):
Looks good! Just a minor detail regarding comment

webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
jorgenherje, line 323:
```suggestion
                    # Configure line width and color scaling to easier separate
                    # statistics traces and realization traces.
                    # Show selected realizations - only filter df if realizations filter
                    # query is not performed
```

webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
jorgenherje, line 332:
```suggestion
                    # Add statistics on top
```

jorgenherje (changes requested):
Suggestions for new comments

anders-kiaer (commented):
Nice - I like it! 👍

jorgenherje (approved):
Looks good! 👍

PR #914, opened 2022-01-07
RftPlotter: update param response selections by clicking in bar chart

lindjoha:
This PR fixes the problem of not being able to update selections by clicking in the bar chart when the chart is in param_vs_sim mode. The trick is to avoid callbacks triggering each other in an infinite loop, but that is solved by updating the whole dropdown objects (for date and zone) instead of only the value.

Also fixes a bug in the callback that was updating the formations well by clicking in the map.


tnatt (approved):
LGTM👍

Issue #920, opened 2022-01-17
Historical vector assume realization = 0 in NewSimulationTimeSeries plugin


jorgenherje:
Historical vectors for `NewSimulationTimeSeries`-plugin are retrieved by use of a single realization number. The algorithm utilize hard coded realization number 0. This results in crash when provider does not contain realization number 0.

`webviz_subsurface/plugins/_simulation_time_series/utils/history_vectors.py`

**Describe the bug**
When dataset/provider does not contain realization number = 0, the provider crash!

**Expected behavior**
Retrieve first possible realization number for provider, and use the historical vector data.

**Fix**
Utilize provider realizations() accessor function and request first element of returned list.

jorgenherje:
Closed by PR https://github.com/equinor/webviz-subsurface/pull/921

PR #921, opened 2022-01-17
Use lowest valid provider realization for History Vectors in new `SimulationTimeSeries`

jorgenherje:
Replace usage of realization = 0, with retrieving valid realization numbers for provider and use lowest realization number to create history vector for new `SimulationTimeSeries`-plugin.

Resolves issue: https://github.com/equinor/webviz-subsurface/issues/920


webviz_subsurface/plugins/_simulation_time_series/utils/history_vectors.py
```diff
@@ -54,13 +54,15 @@ def create_history_vectors_df(
         if historical_vector_name and historical_vector_name in provider.vector_names():
             historical_vector_and_vector_name_dict[historical_vector_name] = vector
 
-    if not historical_vector_and_vector_name_dict:
+    # Get lowest valid realization number
+    realization = (
+        None if len(provider.realizations()) <= 0 else min(provider.realizations())
+    )
```
anders-kiaer, line 60: Is `realizations()` returning a list, or something list-like? If so maybe we can simplfy to this:
```suggestion
    realization = min(provider.realizations(), default=None)
```

anders-kiaer (changes requested):
One comment only, and perhaps add a short changelog entry since it is a bug fix on something previously released?

webviz_subsurface/plugins/_simulation_time_series/utils/history_vectors.py
jorgenherje: Absolutely. Way cleaner code!

anders-kiaer (approved):

PR #924, opened 2022-01-17
Bugfixes and improvements to `ParameterFilter` and `ParameterAnalysis`

tnatt:
PR with bugfixes and improvements to the `ParameterFilter` component and to the `ParameterAnalysis` plugin

### ParameterFilter
**Bugfixes:**
 - Handle NaN values in parameter dataframe that enters the `ParameterFilter`. NaN values in the Select component caused issues when trying to render the layout. 
 
**Added functionality:**
 - Option to choose which ensemble to filter on 
 - Option to use range from selected ensemble or from all ensembles
 - Active Filter indicator where the user can see which parameters are active and their ranges
 - Information regarding which realizations are removed with the current filter   
 - Apply button changes color when selected filters is not in sync with stored filters
 
image

### ParameterAnalysis
**Bugfixes:**
- If only one realization left after filtering the correlation would fail in the `ParameterAnalysis`
- remove sensitivity ensembles as options in the Correlation tab
- fix number formatting in table
- fix issue with persistence to remember selection also when page is refreshed

**Added functionality:**
- boxplot added as option in the parameter distribution tab
- available parameters in both tabs now updates based on selected ensembles 
- update ensemble in the `ParameterFilter ` when ensemble is changed

### DIFF
--- a/webviz_subsurface/_components/parameter_filter.py
+++ b/webviz_subsurface/_components/parameter_filter.py
@@ -1,159 +1,161 @@
 from typing import Dict, List, Tuple
+from uuid import uuid4
 
-import dash_bootstrap_components as dbc
 import numpy as np
 import pandas as pd
 import webviz_core_components as wcc
-from dash import ALL, Dash, Input, Output, State, dcc, html
+from dash import (
+    ALL,
+    Input,
+    Output,
+    State,
+    callback,
+    callback_context,
+    dcc,
+    html,
+    no_update,
+)
 from dash.exceptions import PreventUpdate
-from webviz_config.utils import calculate_slider_step
+
+from webviz_subsurface._models.parameter_model import ParametersModel
+from webviz_subsurface._utils.formatting import printable_int_list
 
 
 class ParameterFilter:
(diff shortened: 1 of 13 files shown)


webviz_subsurface/_components/parameter_filter.py
```diff
@@ -164,111 +166,399 @@ def _initial_store(self) -> Dict[str, List]:
...
-        ),
-        value=[values.min(), values.max()],
-        marks={
-            str(values.min()): {"label": f"{values.min():.2f}"},
-            str(values.max()): {"label": f"{values.max():.2f}"},
+        id={
+            "id": uuid,
+            "type": "slider",
+            "name": name,
+            "r": str(random.random()),  # needed to trigger update (bug)
```
HansKallekleiv, line 518: try `uuid4` instead for the security issue

tnatt: good suggestion, I will 👍

HansKallekleiv (approved):

PR #929, opened 2022-01-19
Don't skip SENSNAME ref if there multiple realizations in tornado

### LINKED ISSUE #928: Don't skip `SENSNAME="ref"` from tornado bars if multiple realizations
The attached code skips all realizations that have `SENSNAME="ref"`, but it might be that you have created a reference set of multiple realizations that you compare with the different scenarios. It is hard to debug the why the span of your "ref-case" has disappeared. 
A more natural behavior would be that "ref" is only skipped if there is only one matching realization. 

https://github.com/equinor/webviz-subsurface/blob/10b792bed9ccc948095855582f102ee44a4c79e3/webviz_subsurface/_components/tornado/_tornado_data.py#L110-L114

asnyv:
### Contributor checklist

- [X] :tada: This PR closes #928.

### DIFF
--- a/webviz_subsurface/_components/tornado/_tornado_data.py
+++ b/webviz_subsurface/_components/tornado/_tornado_data.py
@@ -107,7 +107,7 @@ class TornadoData:
         avg_per_sensitivity = []
 
         for sens_name, sens_name_df in dframe.groupby(["SENSNAME"]):
-            # Excluding cases if ref` is used as `SENSNAME`, and only one realization
+            # Excluding cases if `ref` is used as `SENSNAME`, and only one realization
             # is present for this `SENSNAME`
             if sens_name == "ref" and len(sens_name_df["REAL"].unique()) == 1:
                 continue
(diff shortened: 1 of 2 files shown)


webviz_subsurface/_components/tornado/_tornado_data.py
```diff
@@ -107,10 +107,9 @@ def _calculate_sensitivity_averages(
         avg_per_sensitivity = []
 
         for sens_name, sens_name_df in dframe.groupby(["SENSNAME"]):
-            # Excluding the reference case as well as any cases named `ref`
-            # `ref` is used as `SENSNAME`, typically for a single realization only,
-            # when no seed uncertainty is used
-            if sens_name == "ref":
+            # Excluding cases if ref` is used as `SENSNAME`, and only one realization
```
anders-kiaer, line 110:
```suggestion
            # Excluding cases if `ref` is used as `SENSNAME`, and only one realization
```

anders-kiaer (approved):
LGTM!

PR #932, opened 2022-01-20
`RftPlotter` fix: Handle correlate function returning `NaN` values

lindjoha:
Fixes a small bug related to when the correlate function is returning only `NaN` values. 

Renames the main plugin file to _plugin.py following the best practice and a few other small things.


anders-kiaer (approved):
Add changelog, otherwise LGTM! 🙂

PR #934, opened 2022-01-20
Fix: correct labels in tornado when both cases are on same side of ref

### LINKED ISSUE #933: Error in case labels in Tornado if not of type mc and both high cases on same side of ref
Error in case labels in Tornado if not of type mc and both high cases on same side of ref
**Expected behavior**
Should be "`Case: some_label | the_other_label`"

**Screenshots**
Example from Drogon demo:
image

**Additional context**
The label is hardcoded here, instead of using actual label: 
https://github.com/equinor/webviz-subsurface/blob/39db9b92ad17f0ca0af27b15f8d674d2ab9c8108/webviz_subsurface/_components/tornado/_tornado_bar_chart.py#L110

asnyv:
---

### Contributor checklist

- [X] :tada: This PR closes #933 .

think it isn't necessary with a changelog entry for this.


anders-kiaer (approved):
👌🏻🙂

Issue #908, opened 2022-01-05
Only selected labels shown in `ParameterResponseCorrelation`


anders-kiaer:
...when there are many sensitivities like in this screenshot: 
![image](https://user-images.githubusercontent.com/31612826/148229317-5b6443b8-3367-4aa5-8b55-7cc05bfa9bf6.png)

Currently only e.g. every second label is plotted by Plotly.

Should we as a first simple improvement use `layout.yaxis.dtick = 1` to enforce all labels? Overcrowded amount of labels visually is better than skipping labels (which could lead to misunderstandings). A second improvement could e.g. be to increase height of plot depending on number of sensitivities (with or without a scroll bar frame), or only show top X sensitivities.

anders-kiaer:
Similar to https://github.com/equinor/webviz-subsurface/pull/302

PR #936, opened 2022-01-21
`ParameterResponseCorrelation`: Additional options in GUI

### LINKED ISSUE #465: KeyError when having different parameters in different ensembles in parameter_correlation
Parameter dataframes are stored per ensemble in `ParameterCorrelation`, and each ensemble may not have the same parameters. As the parameter option list is the union of all ensembles parameters, we introduce a risk of KeyError.

Should handle this in a softer way. Best would be "intelligent" handling of parameter options and ensembles to compare with, or at least catch the KeyError and inform why this combination does not work.

### LINKED ISSUE #803: ParameterResponseCorrelation: more options in gui
I would like more of the options in the config file be available in the gui: f.ex correlation method, filters, aggregation method.

Also it should be possible to filter out parameters with correlations below a certain threshold.

### LINKED ISSUE #909: Add parameterfilter to `ParameterResponseCorrelation`
In some cases parameters may be known to be irrelevant for a given response. Any correlation between such pairs could be confusing.
There should be an option to filter parameters (both upfront and interactively).

asnyv:
Draft solution for multiple minor issues
---

### Contributor checklist

- [X] :tada: Relevant issues:
  - closes #909 
  - closes #803 
  - closes #465
  - minimal fix (force all ticks) of #908

- [ ] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR. **-No tests added**


lindjoha (approved):
LGTM

Issue #931, opened 2022-01-20
Fix hover info for observation trace in `SimulationTimeSeries`


jorgenherje:
Fix incorrect hover info for observation trace in `SimulationTimeSeries`-plugin when grouped by `Time Series`. 

Hover info shows: `"Observation: Observation"`:

![image](https://user-images.githubusercontent.com/82032112/150296709-b14fb410-56e3-4888-b9e1-6856fff4b768.png)

**Fix**
Add default behavior:  `legend_group: "Observation"` for `create_vector_observation_traces()`-method, and remove usage of `legend_group` argument in `vector_subplot_builder.py`.

jorgenherje:
Closed by: https://github.com/equinor/webviz-subsurface/pull/937

PR #937, opened 2022-01-24
Fix hover info for observation trace new SimulationTimeSeries plugin

jorgenherje:
Fix incorrect hover info for observation trace in `SimulationTimeSeries`-plugin when grouped by Time Series.

Default name and legend group fixing issue with trace name in hover info.

Resolves issue: https://github.com/equinor/webviz-subsurface/issues/931

----

### Contributor checklist

- Default legend group and name "Observation".
- Updated unit test

### DIFF
--- a/webviz_subsurface/plugins/_simulation_time_series/utils/create_vector_traces_utils.py
+++ b/webviz_subsurface/plugins/_simulation_time_series/utils/create_vector_traces_utils.py
@@ -22,43 +22,50 @@ from ..utils.from_timeseries_cumulatives import is_interval_or_average_vector
 
 
 def create_vector_observation_traces(
-    vector_observations: dict, color: str = "black", legend_group: Optional[str] = None
+    vector_observations: dict,
+    color: str = "black",
+    legend_group: Optional[str] = None,
+    show_legend: bool = False,
 ) -> List[dict]:
     """Create list of observations traces from vector observations
 
     `Input:`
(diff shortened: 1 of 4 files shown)


webviz_subsurface/plugins/_simulation_time_series/utils/create_vector_traces_utils.py
```diff
@@ -22,43 +22,50 @@
...
     List of marker traces for each observation for vector
     """
     observation_traces: List[dict] = []
 
+    _name = "Observation" if legend_group is None else "Observation: " + legend_group
+    _legend_group = "Observation" if legend_group is None else legend_group
+
     for observation in vector_observations.get("observations", []):
-        hovertext = observation.get("comment", "")
+        hovertext = observation.get("comment", None)
```
anderskiaer, line 49:
```suggestion
        hovertext = observation.get("comment")
```
I think `None` is default?

anderskiaer (approved):
LGTM! A minor comment only. Add a changelog entry?

anders-kiaer (approved):

webviz_subsurface/plugins/_simulation_time_series/utils/create_vector_traces_utils.py
jorgenherje: That's true!

PR #944, opened 2022-01-28
`WellCompletion` handling different zone → layer mappings over realizations

### LINKED ISSUE #899: Make `WellCompletion` handle different zone->layer mappings per realization
The `WellCompletion` plugin is dependent on a fixed zone->layer mapping in all realizations. But this is not always the case, f.ex in Drogon the zone layer mapping is varying between realizations. The `zone_layer_mapping.lyr` file is already genereated per realizations, and the solution is that this also needs to be read and used per realization.

lindjoha:
This PR extends the functionality in the `WellCompletion` plugin so that it handles zone->layer mappings that are potentially different across realizations (as is the case in the Drogon sample data).

---

### Contributor checklist

- [x] :tada: This PR closes #899.
   - [x] Import the zone->layer mappings for all realizations and return the data as a dataframe
   - [x] merge the zone->layer dataframe with the compdat data on layer and realization
- [x] :book: I have added an entry in `CHANGELOG.md`

### DIFF
--- a/webviz_subsurface/_datainput/well_completions.py
+++ b/webviz_subsurface/_datainput/well_completions.py
@@ -3,7 +3,7 @@ import json
 import logging
 import re
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional
 
 import pandas as pd
 from ecl2df import EclFiles, common
--- a/webviz_subsurface/_datainput/well_completions.py
+++ b/webviz_subsurface/_datainput/well_completions.py
@@ -41,6 +41,9 @@ def read_zone_layer_mapping(
     df_files = ens.find_files(zone_layer_mapping_file)
     dataframes = []
 
+    if df_files.empty:
+        return pd.DataFrame()
+
     for _, row in df_files.iterrows():
         real = row["REAL"]
         zonelist = common.parse_lyrfile(filename=row["FULLPATH"])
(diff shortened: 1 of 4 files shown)


webviz_subsurface/_datainput/well_completions.py
```diff
@@ -29,22 +30,39 @@ def remove_invalid_colors(zonelist: List[Dict[str, Any]]) -> List[Dict[str, Any]
 
 def read_zone_layer_mapping(
     ensemble_path: str, zone_layer_mapping_file: str
-) -> Tuple[Optional[Dict[int, str]], Optional[Dict[str, str]]]:
+) -> pd.DataFrame:
     """Searches for a zone layer mapping file (lyr format) on the scratch disk. \
     If one file is found it is parsed using functionality from the ecl2df \
     library.
+
+    UPDATE!
```
anders-kiaer, line 38: What does this indicate? Something to do in this PR, or create a GitHub issue?

anders-kiaer (changes requested):
Nice work @lindjoha :clap: Worth a changelog entry! :champagne:

webviz_subsurface/_datainput/well_completions.py
lindjoha: It was a reminder to myself to fix the docstring! I'll fix it

anders-kiaer (approved):

PR #950, opened 2022-02-07
Revised EnsembleSummaryProvider interface

sigurdp:

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/_providers/ensemble_summary_provider/ensemble_summary_provider.py
+++ b/webviz_subsurface/_providers/ensemble_summary_provider/ensemble_summary_provider.py
@@ -22,6 +22,32 @@ class Frequency(Enum):
             return None
 
 
+class DateSpan(Enum):
+    INTERSECTION = "intersection"
+    UNION = "union"
+
+
+@dataclass(frozen=True)
+class ResamplingOptions:
+    """Specifies resampling options, most notably the resampling frequency.
+    Can also specify a `common_date_span` that will influence which dates get included
+    in the returned time series data.
+      * DateSpan.INTERSECTION - truncates the returned range of dates so that all returned
+        realizations have the same date range. The returned range of dates will be
+        the intersection of the date range available per requested realization.
+      * DateSpan.UNION - extends the returned range of dates so that all realizations
+        have same date range and the same dates. The returned range of dates will be the
+        union of the date ranges available per requested realization. Vector values will
+        be extrapolated.
+      * None - each returned realization will contain dates according to the requested
+        frequency, but no effort will be made to truncate or expand the total date range
+        in order to align stert and end dates between realizations.
+    """
+
+    frequency: Frequency
+    common_date_span: Optional[DateSpan] = None
+
+
 @dataclass(frozen=True)
 class VectorMetadata:
     unit: str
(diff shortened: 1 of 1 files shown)


webviz_subsurface/plugins/_simulation_time_series/utils/history_vectors.py
```diff
@@ -60,7 +64,14 @@ def create_history_vectors_df(
         return pd.DataFrame()
 
     historical_vector_names = list(historical_vector_and_vector_name_dict.keys())
+
+    # NOTE: DateSpane = None
```
anders-kiaer, line 68:
```suggestion
    # NOTE: DateSpan = None
```

webviz_subsurface/plugins/_simulation_time_series/types/derived_ensemble_vectors_accessor_impl.py
```diff
@@ -63,10 +67,17 @@ def __init__(
...
-        self._resampling_frequency = (
-            resampling_frequency if self._provider.supports_resampling() else None
-        )
+
         self._relative_date = relative_date
+        self._resampling_options = (
+            resampling_options if self._provider.supports_resampling() else None
+        )
+
+        # Make date span union if calculating data relative to date
```
anders-kiaer, line 76: Should we expand on this comment? Not immediately clear to me why data relative to date → we should take date unione.

anders-kiaer (commented):
:heart: the tests! :robot: Great code style as well - on a difficult domain technical issue. :clap: 

I'm not able to find much to comment on, but two questions:

1) The pattern 
   ```python
   ResamplingOptions(resampling_frequency, date_span)
   if resampling_frequency
   else None
   ```
   appears to repeat itself somewhat. Should we consider adjusting `ResamplingOptions` to have type `frequency: Optional[Frequency]` such that a `frequency = None` can be passed directly, similar to `common_date_span` also being `Optional` (i.e. support `None`)? Or maybe this change is not relevant anymore after the second question:

2. I'm not 100% sure that I remember correctly our conclusion from February, but what I _think_ I remember is that expected logic behind the scenes (as end user), wrt. how date is handled (intersection, union or raw), is along the lines of:
   - When only individual realizations are plotted, we don't need to care about union/intersection of dates. Each realization can if we want be plotted with its own date span. 
   - When any kind of statistics is calculated, we need to either take union or intersection across the dates (to get a common date axis). For vectors that are of type _cumulative_ we can take the union (since these have a well defined extrapolation). For all other types of vectors we take the intersection (since these are unsafe to extrapolate - we could potentially also extrapolate rate vectors by assuming they are zero outside data span, but I'm not sure if we today have the metadata necessary to differentiate between rate vectors and e.g. pressure vectors? @asnyv?).

If this is the logic we want, resampling frequency is (still) a general property across all vectors accessed, but the date span (union/intersection) depends on vector type. :thinking: 

Do I recall incorrectly, or has something else popped up during development that forces us to reconsider this logic? Or maybe something along these lines is already implemented (and I'm just reading the code incorrectly). :laughing: :slightly_smiling_face:

asnyv:
> When any kind of statistics is calculated, we need to either take union or intersection across the dates (to get a common date axis). For vectors that are of type cumulative we can take the union (since these have a well defined extrapolation). For all other types of vectors we take the intersection (since these are unsafe to extrapolate - we could potentially also extrapolate rate vectors by assuming they are zero outside data span, but I'm not sure if we today have the metadata necessary to differentiate between rate vectors and e.g. pressure vectors? @asnyv?).

@anders-kiaer We don't have the metadata, but I think maybe a varying date span between statistics of different vectors could be a bit confusing for users, and that we should consider to go for the union for all vectors?

PR #940, opened 2022-01-27
User defined vector definition in SimulationTimeSeries plugin

jorgenherje:
Add user defined vector definitions for `SimulationTimeSeries`- plugin

---
- Add configurable user defined vector definitions for plugin 
- Rename `AVG_` and `INVTL_` to `PER_DAY_` and `PER_INTVL_`, respectively. For vector it is calculated from
    - **NOTE**: Does not replace `T`-suffix with `R`-suffix for `PER_DAY_`-vector
    - Previously `WOPT` gave `AVG_WOPR` and `INTVL_WOPT`. Not it gives `PER_DAY_WOPT` and `PER_INTVL_WOPT`
- Remove `webviz_subsurface/_abbreviations/abbreviation_data/reservoir_simulation_vectors.json` and retrieve `VectorDefinitions` using new in Python-API for `webviz-subsurface-components`.
---

## NOTE:
Resolves:
* https://github.com/equinor/webviz-subsurface/issues/321

Dependent on: 
* https://github.com/equinor/webviz-subsurface-components/pull/782
* https://github.com/equinor/webviz-subsurface-testdata/pull/126

PR #951, opened 2022-02-08
Delta relative to date within ensemble `SimulationTimeSeries`

jorgenherje:
Calculate delta relative to date within ensemble for `SimulationTimeSeries`-plugin

---

### Contributor checklist
* Selectable dates in dropdown - union of dates from all `EnsembleSummaryProvider`'s in provider set
* Calculate vector data in dataframes relative to selected date - if date does not exist -> return empty df
* Calculate statistics afterwards, i.e. "resetting" the statistics calculation!
* Assume set of realizations to exist for each date in an ensemble, i.e. all realizations exist for all dates in an ensemble. Will need updated provider-interface.

### NOTE:
* Should await merge of PR: https://github.com/equinor/webviz-subsurface/pull/940 - as it is branched
* Coming modification of `EnsembleSummaryProvider`-interface in PR: https://github.com/equinor/webviz-subsurface/pull/950


webviz_subsurface/plugins/_simulation_time_series/utils/dataframe_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        columns=["DATE"]
+    )
+    if _relative_date_df.empty:
+        # Dataframe with columns, but no rows
+        return output_df
+
+    vectors = [elm for elm in df.columns if elm not in ("DATE", "REAL")]
+
+    # NOTE: This for-loop makes it possible to get real not represented in _relative_date_df!
+    for _real, _df in df.groupby("REAL"):
```
anders-kiaer, line 55:
```suggestion
    for real, df in df.groupby("REAL"):
```
Or is there a reason behind the leading `_`? I think in other place we use leading `_` to indicate an unused variable (a practice recognized by e.g. `pylint`), but here they are used in the code block.

webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
```diff
@@ -561,6 +593,10 @@ def _user_download_data(
 
             # Append data for each vector
             for vectors_df in vectors_df_list:
+                # Ensure rows of data
+                if vectors_df.shape[0] <= 0:
```
anders-kiaer, line 597:
```suggestion
                if not vectors_df.shape[0]:
```

webviz_subsurface/plugins/_simulation_time_series/_callbacks.py
```diff
@@ -308,6 +323,10 @@ def _update_graph(
                 )
 
             for vectors_df in vectors_df_list:
+                # Ensure rows of data
+                if vectors_df.shape[0] <= 0:
```
anders-kiaer, line 327:
```suggestion
                if not vectors_df.shape[0]:
```

anders-kiaer (changes requested):
A couple of small comments, otherwise LGTM! :clap: :rocket:

anders-kiaer (approved):

PR #958, opened 2022-02-13
Disable rangeslider marks in cross-section plugins

HansKallekleiv:
In Dash > 2.10 the `RangeSlider` component automatically calculates `marks` from `min` and `max`.
If `min` and `max` is not provided this calculation fails.
Marks are not needed for our use, so we can set `marks=None` to disable this behaviour.

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


anders-kiaer (approved):
👍

PR #965, opened 2022-02-23
Allow filtered subset of surface names for multiple attributes in `StructuralUncertainty`

HansKallekleiv:
Relates to #962


tnatt (approved):
LGTM

PR #969, opened 2022-02-23
SimulationTimeSeries: Change default options from None to empty dict

### LINKED ISSUE #968: Bug: option deprecation check in SimulationTimeSeries fails if options is not defined
This (currently unreleased) check requires that `options` is an `Iterable`, but the default  value is `None`
https://github.com/equinor/webviz-subsurface/blob/4f7d31d8c95cfda9dd6fd2713bb16404a1d3e4e4/webviz_subsurface/plugins/_simulation_time_series/_plugin.py#L51-L61

Suggestion: Either skip if `options is None` or change default for options to an empty list `[]`

asnyv:

---

### Contributor checklist

- [X] :tada: This PR closes #968 .
    **_Skipped as the bug has not been released._**


asnyv:
Was also able to remove the `plot_options` as it was only there to handle the `options is None` scenario and set it to an empty dict.

asnyv:
Seems like `pylint` dislikes empty dicts as default value @jorgenherje

jorgenherje:
I guess `None` and early return on `!options` in `check_deprecation_argument()` is better? As empty `dict` is a defined object without content. Not defining `options` would be more representable with `None`

webviz_subsurface/plugins/_simulation_time_series/_plugin.py
```diff
@@ -267,7 +269,7 @@ def __init__(
             self._initial_vector_selector_data, self._predefined_expressions
         )
 
-        plot_options = options if options else {}
+        plot_options = {} if options is None else options
```
asnyv, line 272: see that I was a bit fast and rewrote it differently, will change back. No need to change this line 😅

PR #970, opened 2022-02-23
Replace modals with `wcc.Dialog` in `StructuralUncertainty`

### LINKED ISSUE #967: StructuralUncertainty: Use `wcc.Dialog` instead of modals

HansKallekleiv:
- Replace modals with new Dialog component
- Rewrote uncertainty table to display as a Dialog.

Closes #967


webviz_subsurface/plugins/_structural_uncertainty/structural_uncertainty.py
```diff
@@ -188,28 +186,30 @@ def __init__(
             ens: webviz_settings.shared_settings["scratch_ensembles"][ens]
             for ens in ensembles
         }
+
+        # Create a table of surface files
         surface_table = find_surfaces(self._ensemble_paths)
+        # Filter on provided surface attributes
         surface_table = surface_table[surface_table["attribute"].isin(self._surf_attrs)]
+        # Filter on provided surface names
+        self._surfacenames = (
```
tnatt, line 195: Question, I have not tested yet.. but what happens if the user has a typo in one of the names and it does not exist, will the plugin handle that? should we check against the available names and warn if the user has given an invalid name?

webviz_subsurface/plugins/_structural_uncertainty/views/dialog.py
```diff
@@ -0,0 +1,61 @@
...
+        ),
+    )
+
+
+def clear_all_apply_dialog_buttons(
+    uuid: str, dialog_id: str, apply_disabled: bool = True
+) -> html.Div:
+    return html.Div(
+        children=[
+            dbc.Button(
```
tnatt, line 40: These buttons can be changed into regular html.Buttons to remove the bootstrap dependency here as well. I did that in the ParameterFilter and the buttons looked and behaved the same

tnatt (approved):
LGTM, two comments/questions only 🙂

webviz_subsurface/plugins/_structural_uncertainty/views/dialog.py
HansKallekleiv: Thanks. Will fix in a separate pr :) #974

PR #972, opened 2022-02-24
Two bugfixes in `VolumetricAnalysis`

tnatt:
Bugfixes in `VolumetricAnalysis`:

-  Selecting only BO and/or BG in the table introduced a bug in pandas saying "no numeric types to aggregate". This is avoided by setting these to nan instead of "NA" when the data was not grouped on "FLUID_ZONE"  
- If volumefiles from one ensemble contained PORE and volumefiles from another enemble contained PORV, the PORV renaming led to two columns with PORV in the merged dataset and an error occured.

---

### Contributor checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


HansKallekleiv (approved):
LGTM :)

PR #980, opened 2022-03-10
Deprecate `SurfaceViewerFMU`

HansKallekleiv:
Deprecate `SurfaceViewerFMU` and point user to the new `MapViewerFMU` plugin.


webviz_subsurface/plugins/_surface_viewer_fmu.py
```diff
@@ -10,6 +10,7 @@
 from dash.exceptions import PreventUpdate
 from webviz_config import WebvizPluginABC, WebvizSettings
 from webviz_config.webviz_store import webvizstore
+from webviz_config.deprecation_decorators import deprecated_plugin
```
anders-kiaer, line 13: The import order linting is complaining about this one. Reorder and then we merge? 🎉

anders-kiaer (changes requested):

anders-kiaer:
Also add a changelog entry stating we deprecate `SurfaceViewer`?

anders-kiaer (approved):

PR #988, opened 2022-03-25
Arrow support and other improvements to `ParameterAnalysis` and `PropertyStatistics`

tnatt:
PR includes improvements and some bugfixes to the `ParameterAnalysis` and `PropertyStatistics` plugins (these plugins are very similar) 

### Both Plugins
- Use arrow format for unsmry 
- Deprecation of the use of aggregated csvfiles
- Refactoring to use more common functions
- `VectorSelector` implemented

### ParameterAnalysis
- Added option of showing statistical lines #982 
- Several performance enhancing measures
  - improved speed of generating the timeseries figure
  - do not correlate the selected parameter against all vectors as default, but base it on a user selected column key filter
- use wcc.dialog for the options
- increase precision of parameter table

### PropertyStatistics
- Added scatter figure to show the correlation between the property and the selected vector
- Bugfix handle missing data after filtering
---


HansKallekleiv (approved):
LGTM 🥇

Issue #993, opened 2022-03-29
License Boundary Plugin


vegardkv:
The goal of this plugin is to provide insight into potential issues with CO2 migrating outside a license boundary. One of the main components of the plugin will be a map visualizing the CO2 plume, license boundary, fault polygons, well paths/picks and possibly other elements. In addition, there will be components illustrating which realizations that has issues with CO2 leakage outside the license boundary, and some indicator of severity for the issue.

The code structure of the MapViewerFMU plugin will most likely be used for inspiration.

Issue #99, opened 2019-09-30
Remove matplotlib dependency?


anders-kiaer:
Currently `matplotlib` is a dependency. It is only used [for getting some preselected colormaps](https://github.com/equinor/webviz-subsurface/blob/0f8d82cca172a4f6b3fe6021f2c9a097757631cf/webviz_subsurface/datainput/layeredmap/_image_processing.py#L5). Instead add colormap definitions to `webviz-subsurface` and remove the quite big `matplotlib` dependency?

anders-kiaer:
Closed by #420.

PR #994, opened 2022-03-29
CO₂ migration plugin

vegardkv:
Aimed at #993 

Implements a new plugin for CO2 leakage analysis, using the new layout framework.

Example data is available via https://github.com/equinor/webviz-subsurface-testdata/issues/136

(Known) remaining tasks:
- [x] Code formatting


webviz_subsurface/plugins/_co2_leakage/_utilities/co2volume.py
```diff
@@ -0,0 +1,136 @@
...
+        df,
+        y=_Columns.REALIZATION.value,
+        x=_Columns.VOLUME.value,
+        color=_Columns.CONTAINMENT.value,
+        title="End-state CO2 containment",
+        orientation="h",
+        category_orders={_Columns.CONTAINMENT.value: ["outside", "inside"]},
+        color_discrete_sequence=["#dd4300", "#006ddd"],
+    )
+    fig.layout.height = height
```
HansKallekleiv, line 73: Not sure if it is intended, but these size attributes generates somewhat small figures.
image

Without:
image

You can also consider replacing `dcc.Graph` with the `wcc.Graph` wrapper which will add responsive resizing.
https://github.com/equinor/webviz-core-components/blob/master/webviz_core_components/wrapped_components/graph.py

webviz_subsurface/plugins/_co2_leakage/_utilities/surface_publishing.py
```diff
@@ -0,0 +1,104 @@
...
+    StatisticalSurfaceAddress,
+    SimulatedSurfaceAddress,
+)
+from webviz_subsurface._providers.ensemble_surface_provider.ensemble_surface_provider import \
+    SurfaceStatistic
+from webviz_subsurface.plugins._co2_leakage._utilities.plume_extent import truncate_surfaces
+
+
+@dataclass
+class TruncatedSurfaceAddress:
```
HansKallekleiv, line 20: This is interesting and should probably be seen in a larger context on how to handle calculated surfaces.
I guess it also raises the question if the current FMU standards for surface naming are "rich" enough.
Nice workaround :)

webviz_subsurface/plugins/_co2_leakage/views/mainview/settings.py
```diff
@@ -0,0 +1,367 @@
...
+            return len(min_auto) == 1, len(max_auto) == 1
+
+
+class FilterSelectorLayout(wcc.Selectors):
+    def __init__(self, formation_id):
+        super().__init__(
+            label="Filter Settings",
+            children=[
+                "Formation",
+                wcc.Dropdown(
```
HansKallekleiv, line 137: Should this be clearable?

webviz_subsurface/plugins/_co2_leakage/views/mainview/settings.py
```diff
@@ -0,0 +1,367 @@
...
+                    id=ensemble_id,
+                    options=[
+                        dict(value=en, label=en)
+                        for en in ensembles
+                    ],
+                    value=ensembles[0],
+                    clearable=False,
+                ),
+                "Realization",
+                wcc.Dropdown(
```
HansKallekleiv, line 288: For multi selection for many realizations consider using `wcc.SelectWithLabel`

webviz_subsurface/plugins/_co2_leakage/views/mainview/mainview.py
```diff
@@ -0,0 +1,185 @@
...
+                    config={
+                        "displayModeBar": False,
+                    }
+                ),
+            ],
+            **kwargs,
+        )
+
+
+def _color_tables():
```
HansKallekleiv, line 127: This is defined in two files. Suggest moving to a separate file.

webviz_subsurface/plugins/_co2_leakage/_plugin.py
```diff
@@ -0,0 +1,298 @@
...
+                rz_paths,
+                ENSEMBLE_PLOT_HEIGHT,
+                ENSEMBLE_PLOT_WIDTH,
+                self._co2_containment_relpath,
+            )
+            fig0 = generate_co2_volume_figure(*fig_args)
+            fig1 = generate_co2_time_containment_figure(*fig_args)
+            return fig0, fig1
+
+        @callback(
```
HansKallekleiv, line 155: Although it works, Dash is complaining about this callback.
```
A nonexistent object was used in an `Output` of a Dash callback. The id of this object is `CO2Leakage-1-main-view-main-element-date-slider` and the property is `marks`. The string ids in the current layout are: [location, webviz-content-manager, main-menu, settings-drawer, plugins-wrapper, store-date-store-CO2Leakage-1, CO2Leakage-1, CO2Leakage-1-main-settings, CO2Leakage-1-main-settings-ensemble, CO2Leakage-1-main-settings-realization, CO2Leakage-1-main-settings-formation, CO2Leakage-1-main-settings-property, CO2Leakage-1-main-settings-statistic, CO2Leakage-1-main-settings-color-scale, CO2Leakage-1-main-settings-cm-min, CO2Leakage-1-main-settings-cm-min-auto, CO2Leakage-1-main-settings-cm-max, CO2Leakage-1-main-settings-cm-max-auto, CO2Leakage-1-main-settings-plume-threshold, CO2Leakage-1-main-settings-plume-smoothing]
```
I'm not entirely sure why, but I think it is related to the use of a store, and that the callbacks are registered on the main plugin before all components are added? 
cc: @rubenthoms

webviz_subsurface/plugins/_co2_leakage/_utilities/generic.py
```diff
@@ -0,0 +1,35 @@
+import pathlib
+from enum import Enum
+from typing import Dict, Iterable, Optional
+
+import numpy as np
+from fmu.ensemble import ScratchEnsemble
+
+
+def fmu_realization_paths(ensemble_path) -> Dict[int, str]:
```
HansKallekleiv, line 9: Any function that works directly with the filesystem must be wrapped with the `webvizstore` decorator or use a provider approach. This is because if the webviz application is made "portable" and moved e.g. to radix, there will no longer be access to the original files.

How to solve it is a longer discussion which we can have in a meeting :)

HansKallekleiv (changes requested):
I have tested with the provided [testdata](https://github.com/equinor/webviz-subsurface-testdata/pull/137), and the plugin looks good!

There are some issues that must be fixed and some that must be discussed :)

webviz_subsurface/plugins/_co2_leakage/_utilities/co2volume.py
vegardkv: Removed fixed height/width and changed to wcc.Graph as suggested. Looks better now!

webviz_subsurface/plugins/_co2_leakage/views/mainview/settings.py
vegardkv: No. Fixed.

webviz_subsurface/plugins/_co2_leakage/views/mainview/mainview.py
vegardkv: Well picked up. Fixed as suggested

webviz_subsurface/plugins/_co2_leakage/_utilities/generic.py
vegardkv: Removed all direct usage of the two FMU path-related functions in this file. The functionality was easily replaced by using ´read_csv´ from `webvizstore_functions`. I believe this resolves the issue properly.

Well picks and the boundary polygons are now dealt with similarly to how well picks are handled in MapViewerFMU. I guess the main drawback is that this does not allow for differing well picks or boundary polygons across realizations:
- For well picks, I don't know if this is a concern? If it is, then it should also be solved for MapViewerFMU?
- For the boundary polygon, this may be an issue in case the boundary polygon is derived from other data sources that differ between realizations. E.g. if it is derived from the structural model (faults, horizons, etc.). That being said, at this stage it is not critical to fix.

webviz_subsurface/plugins/_co2_leakage/views/mainview/settings.py
vegardkv: Fixed. Agree that this is more user-friendly.

webviz_subsurface/plugins/_co2_leakage/_utilities/surface_publishing.py
vegardkv: Agree that this is probably a bigger-picture discussion. One way could be to extend `SurfaceServer` to handle qualified addresses in a best-effort manner (when the type is unknown), e.g. using a simpler hashing mechanism for those types of objects.

With regards to meta data, I guess there is some work going on in xtgeoapp_grd3dmaps and fmu-dataio that might relate to this, but I am not familiar with the details.

webviz_subsurface/plugins/_co2_leakage/_plugin.py
vegardkv: Seems probably. Found an approach to get rid of the data store, and this seems to have fixed the issue.

vegardkv:
All comments should now have been addressed. There are some discussion points remaining, but I think they mostly relate to"big-picture" issues.

vegardkv:
Fixed black and pylint reports. Came across an issue related to dependencies. This plugin implements a contour line generator which depends on matplotlib and shapely. I cannot see that these are used elsewhere. What is the policy for new dependencies?

There are some options here:
- Allow them (i.e. do nothing)
- Enable functionality based on package availability. If an ImportError occurs, this specific feature (contour lines) is disabled.
- Remove the functionality (for now). The implementation is experimental and not the most important. Its goal is to generate contour lines, but this can be left to the deck.gl component instead. It seems to have contour features, as well as https://github.com/equinor/webviz-subsurface-components/issues/622. However, I couldn't find a way to do make it work as I wanted.

HansKallekleiv:
> Fixed black and pylint reports. Came across an issue related to dependencies. This plugin implements a contour line generator which depends on matplotlib and shapely. I cannot see that these are used elsewhere. What is the policy for new dependencies?
> 
> There are some options here:
> 
> * Allow them (i.e. do nothing)
> * Enable functionality based on package availability. If an ImportError occurs, this specific feature (contour lines) is disabled.
> * Remove the functionality (for now). The implementation is experimental and not the most important. Its goal is to generate contour lines, but this can be left to the deck.gl component instead. It seems to have contour features, as well as [Contour layer webviz-subsurface-components#622](https://github.com/equinor/webviz-subsurface-components/issues/622). However, I couldn't find a way to do make it work as I wanted.

For Equinor usage (komodo) I don't see this as a problem as both packages are already included.
@anders-kiaer , @asnyv  any thoughts?

asnyv:
Sorry, didn't see this until now.

We have tried before to avoid including unnecessary dependencies, and intentionally removed `matplotlib` as a dependency before due to size vs added value (see: #99). Of course not saying that we cannot add more dependencies, but think in a growing repo with several contributors, we should be a bit conservative. If the functionality is more or less experimental, at least I suggest you go for option 2 or 3 from your suggestions (so either try to import and disable if not available, or drop it). If it is found to be necessary later we can rather include the dependencies then.

vegardkv:
Added a check for matplotlib + shapely availability. If an ImportError is raised, the functionality is effectively turned off. I think this is a reasonable solution for now, as this is not a critical part of the plugin. It is an experimental feature that we might want to improve upon later.

That being said, I suspect matplotlib and shapely are implicit dependencies of webviz_subsurface after all, probably via xtgeo? (I didn't really test properly until after the commit above).  When I try to run this plugin, I get ImportError if matplotlib or shapely is not installed, originating from other parts of the code.

Since matplotlib/shapely is not explicit dependencies, I think the solution implemented in b7764aa should be fine

PR #996, opened 2022-03-29
Bugfix `VolumetricAnalysis` - tornados hidden if both dynamic and static sources

tnatt:
An old statement (not needed anymore) caused the "Tornadoplots" tab in `VolumetricAnalysis` to not appear if both static and dynamic volume sources were mixed.


anders-kiaer (approved):
👍

PR #1001, opened 2022-04-07
Fix issues with just one parameter and for aggregated data in `ParameterResponseCorrelation`

### LINKED ISSUE #1000: ParameterResponseCorrelation not working with aggregated data and/or just one parameter

asnyv:
---

Tbh; Not sure why the parenthesis is needed for the filter selectors, but for some reason it makes a difference...

### Contributor checklist

- [X] :tada: This PR closes #1000.
   - [X] Fix issue with slider being misconfigured. Error only occured if only using 1 parameter
   - [X] Paranthesis for filter selectors for some strange reason needed when using aggregated data. No idea why it helps.

 Have skipped changelog, as I am not really sure why this in fact is a "fix". Seems to work though...

### DIFF
--- a/webviz_subsurface/plugins/_parameter_response_correlation.py
+++ b/webviz_subsurface/plugins/_parameter_response_correlation.py
@@ -308,10 +308,10 @@ folder, to avoid risk of not extracting the right data.
                     id=self.uuid("max-params"),
                     min=1,
                     max=max_params,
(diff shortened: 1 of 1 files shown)


anders-kiaer (approved):

lindjoha (approved):
Looks good! Have been tested locally by @runhol

PR #1012, opened 2022-04-25
Fix `CHANGELOG.md`

anders-kiaer:


rnyb (approved):
Good spotted, thanks.

PR #1013, opened 2022-04-26
Expose surface folder as an argument in `MapViewerFMU`

tnatt:
The surface folder was hard coded to "share/results/maps". This PR exposes the surface folder as an argument for the users.


webviz_subsurface/_providers/ensemble_surface_provider/ensemble_surface_provider_factory.py
```diff
@@ -61,13 +61,19 @@ def instance() -> "EnsembleSurfaceProviderFactory":
         return factory
 
     def create_from_ensemble_surface_files(
-        self, ens_path: str, attribute_filter: List[str] = None
+        self,
+        ens_path: str,
+        attribute_filter: List[str] = None,
+        rel_surface_folder: str = "share/results/maps",
```
anders-kiaer, line 67:
```suggestion
        rel_surface_folder: str,
```
Should we remove the default value here in order to only have the default path value defined in one place (less risk of later introduce inconsistencies)?

tnatt: I agree :slightly_smiling_face:

anders-kiaer (approved):

Issue #858, opened 2021-11-25
Move away from `.UNSMRY` to `.arrow` provider


anders-kiaer:
- [x] `ParameterAnalysis` @tnatt 
- [x] `GroupTree` @lindjoha - #902 
- [x] `PropertyStatistics`
- [x] `BhpQc` @rnyb 
- [x] `ParameterResponseCorrelation` @lindjoha 
- [ ] `ReservoirSimulationTimeSeriesOneByOne` @tnatt and @lindjoha #1217 
- [ ] `ReservoirSimulationTimeseriesRegional` @asnyv 
- [ ] `ParameterParallelCoordinates` @anders-kiaer 
- [ ] `HistoryMatch` - ping ERT-team if they can dump data as input to us (@anders-kiaer)

Don't think these currently support directly from UNSMRY, but supports from csv, and should ideally have summary support. In that way they can be used with the same providers as the others if defined with the same data, avoiding unnecessary overhead
- [ ] `LinePlotter` (currently only supports csv, but common user request to have special support for summary data)
- [ ] `TornadoPlotterFMU`  (Similar to  supports csv, but common user request to have special support for summary data)

PR #1028, opened 2022-05-11
`BhpQc` - Read from `.arrow` files instead of `.UNSMRY` files

rnyb:
Read data from arrow files. Related to issue #858.

Dataframe is still built in init so there may be potential for further improvements, but this should work as a first pass implementation of using arrow files instead of UNSMRY files.


anders-kiaer (changes requested):
+ add a changelog entry, then LGTM! :+1:  :slightly_smiling_face:

anders-kiaer (approved):

PR #1030, opened 2022-05-15
New data provider in `ParameterResponseCorrelation`

lindjoha:
This PR implements the new summary data provider i `ParameterResponseCorrelation`. 

I also needed to make a general function that matches column_keys with * as wildcard. I have put it in EnsembleSummaryProvider, but we can discuss where it should be.

### Contributor checklist

- [x] :tada: This PR closes `ParameterResponseCorrelation `in #858
   - [x] New data provider
   - [x] General function that matches column_keys with wild cards

### DIFF
--- a/webviz_subsurface/_providers/ensemble_summary_provider/utils.py
+++ b/webviz_subsurface/_providers/ensemble_summary_provider/utils.py
@@ -1,0 +1,27 @@
+import fnmatch
+import re
+from typing import List
+
+from .ensemble_summary_provider import EnsembleSummaryProvider
+
+
+def get_matching_vector_names(
+    provider: EnsembleSummaryProvider, column_keys: List[str]
+) -> List[str]:
+    """Returns a list of vectors that match the input columns_keys that
+    can have unix shell wildcards.
+
+    Example of use:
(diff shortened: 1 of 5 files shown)


webviz_subsurface/_providers/ensemble_summary_provider/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    df = provider.get_vectors_df(matching_vector_names, None)
+
+    """
+    try:
+        regex = re.compile(
+            "|".join([fnmatch.translate(col) for col in column_keys]),
+            flags=re.IGNORECASE,
+        )
+        return [vec for vec in provider.vector_names() if regex.fullmatch(vec)]
+    except re.error:
```
anders-kiaer, line 26: For my understanding: Do you have some typical examples of when `re.error` will be raised?

anders-kiaer (commented):
LGTM. Minor comment only.

Should add a changelog entry perhaps as this is a somewhat big change for `ParameterResponseCorrelation`?

anders-kiaer (approved):
Good work 👏 LGTM

PR #1046, opened 2022-06-07
Fix for layer changes in map component

HansKallekleiv:
Fix for changes related to https://github.com/equinor/webviz-subsurface-components/issues/1030.


tnatt (approved):
LGTM

PR #1053, opened 2022-06-09
Fix relative path issue in SwatinitQC

tnatt:
PR to ensure relative paths for the faultlines argument works. In addition added x-axis link for SWAT and SWATINIT plot.

### DIFF
--- a/webviz_subsurface/plugins/_swatinit_qc/_plugin.py
+++ b/webviz_subsurface/plugins/_swatinit_qc/_plugin.py
@@ -42,7 +42,7 @@ or with the "check_swatinit" command line tool.
         csvfile: str = "share/results/tables/check_swatinit.csv",
         ensemble: Optional[str] = None,
         realization: Optional[int] = None,
-        faultlines: Optional[Path] = None,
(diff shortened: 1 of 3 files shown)


webviz_subsurface/plugins/_swatinit_qc/_plugin.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         csvfile: str = "share/results/tables/check_swatinit.csv",
         ensemble: Optional[str] = None,
         realization: Optional[int] = None,
```
rnyb, line 44: Skip Optional? i.e. just use` str  = None`

webviz_subsurface/plugins/_swatinit_qc/_plugin.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         csvfile: str = "share/results/tables/check_swatinit.csv",
         ensemble: Optional[str] = None,
```
rnyb, line 43: Skip Optional here too?

PR #1055, opened 2022-06-10
`ProdMisfit`- fixed bug related to use of excl_name_contains.

rnyb:
It was not taken into account, i.e. it had no effect, due to badly placed continue
inside nested loop.

### DIFF
--- a/webviz_subsurface/plugins/_prod_misfit/_plugin.py
+++ b/webviz_subsurface/plugins/_prod_misfit/_plugin.py
@@ -225,13 +225,11 @@ def _get_wells_vectors_phases(
             continue
 
         well = vector.split(":")[1]
-        if well.startswith(tuple(excl_name_startswith)):
+
+        if _skip_well(well, excl_name_startswith, excl_name_contains):
             drop_list.append(well)
             continue
-        for excl in excl_name_contains:
-            if excl in well:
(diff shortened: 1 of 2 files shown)


webviz_subsurface/plugins/_prod_misfit/_plugin.py
```diff
@@ -255,6 +253,23 @@ def _get_wells_vectors_phases(
...
 
+def _skip_well(
+    well: str,
+    excl_name_startswith: list,
+    excl_name_contains: list,
+) -> bool:
+    """Check well name against exclude strings and return True if it should be skipped."""
+
+    # ensure all list elements are str
+    excl_name_startswith = [str(element) for element in excl_name_startswith]
```
anders-kiaer, line 264:
```suggestion
    excl_name_startswith = [str(element) for element in excl_name_startswith]
```
What other types do we expect than `str`? Can we clean this "further up" and ensure `excl_name_startswith = List[str]` to begin with?

rnyb: int could be expected, though str would be more common. In the case of int input, the user can now use the list form [5, 36] as alternative to ["5", "36"]

anders-kiaer (approved):

PR #1057, opened 2022-06-21
New implementation of `EnsembleTableProvider`

lindjoha:
This is still a draft and I would like some feedback before I continue.

I have re-written the `EnsembleTableProvider` to make it similar to the `EnsembleSummaryProvider`. I have for now removed the possiblity of returning a `EnsembleTableProviderSet`, so it will require changes in some plugins. We can choose to add that functionality, but it might be best not to ? I have also removed the `.parquet`-implementation.

The factory class is completely re-written. The `EnsembleTableProviderImplArrow` is similar to before, but has a new function 
`write_backing_store_from_per_realization_tables()`. 

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2
- [x] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.


HansKallekleiv (approved):
LGTM !
Only two very minor comments on formatting :)

PR #1058, opened 2022-06-27
New implementation of the `WellCompletion` plugin (with WLF).

### LINKED ISSUE #900: Single realization option in `WellCompletion`
The `WellCompletion` plugin is only set up to use an ensemble data set. But it would be useful to also have an option to select single realizations. I think this cannot be done front end since the data set would become to large. Backend it is straight forward to do it, the problem is what to do with the portable option. Webviz store is currently saving the finished data set, but that can't be done for every realisation (that would become too much data). So the webviz-store would have to be moved to the input data. In which case the plugin would be slower when run portable since putting together the data takes some time.

lindjoha:
This is a new implementation of the `WellCompletions` plugin. This is based on the `wellcompletiondata` export from `ecl2df `where layer data is already aggregated to zone. This improves speed significantly and opens up for the single realization option.

This is also based on a new implementation of the `EnsembleTableProvider` which is in PR #1057. And it is built in the new WLF framework. 

---

### Contributor checklist

- [x] :tada: This PR closes #900  and #1064
   - [x] Import well completion arrow files with EnsembleTableProvider
   - [x] Write new functionality for creating the dataset for the WellCompletion component
   - [ ] Get KH unit from arrow file metadata


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED] - YYYY-MM-DD
 
+### Added
+
+- [#1058](https://github.com/equinor/webviz-subsurface/pull/1058) - `WellCompletion` - New implementation of the WellCompletions plugin, which is faster, has more functionality (single realization) and utilizes the webviz layout framework (WLF).
```
anders-kiaer, line 12:
```suggestion
- [#1058](https://github.com/equinor/webviz-subsurface/pull/1058) - `WellCompletion` - New implementation of the `WellCompletions` plugin, which is faster, has more functionality (single realization) and utilizes the webviz layout framework (WLF).
```

webviz_subsurface/plugins/_well_completion/views/_well_completion_view.py
```diff
@@ -0,0 +1,206 @@
+from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple
+
+# from dash.development.base_components import Component
```
anders-kiaer, line 4:
```suggestion
```

webviz_subsurface/plugins/_well_completions/_plugin.py
```diff
@@ -2,12 +2,18 @@
...
 from webviz_config import WebvizPluginABC, WebvizSettings
+from webviz_config.deprecation_decorators import deprecated_plugin
 
 from ._business_logic import WellCompletionsDataModel
 from ._callbacks import plugin_callbacks
 from ._layout import layout_tour_steps, main_layout
 
 
+@deprecated_plugin(
+    "This plugin has been replaced by the `WellCompletion` plugin "
```
anders-kiaer, line 13: Should we emphasize that the difference (wrt. plugin name) is the removed `s`? Might be hard to see the difference immediately for the users.

anders-kiaer (changes requested):
Very impressive code quality @lindjoha! 👏 This new implementation of the plugin will also be a nice improvement. Some very small comments, then we're ready for merge IMO.

webviz_subsurface/plugins/_well_completions/_plugin.py
lindjoha: Good idea :)

anders-kiaer (approved):

anders-kiaer (approved):
🎉

PR #1097, opened 2022-08-03
removed requirement of "" around wildcarded paths in smry2arrow_batch

### LINKED ISSUE #1096: `smry2arrow_batch` allow for not using "" around wildcarded paths
The smry2arrow_batch script has become quite popular and reached a state where I think it might have to be considered as a more or less permanent tool rather than a temporary solution, both as users sometimes run Eclipse outside ERT, and because many workflows still lack the ECL2CSV job with the arrow option. A small issue is though the requirement of "" around the wildcarded path, which users often forget/ignore. 
 
If not using "" the unix command line (at least in e.g. bash and cshell) will prosess the wildcards like * and give a bunch of inputs to the script, whilst we only use the first one, and then use glob to find multiple paths for the ensemble. I suggest that we also support an undefined number of input paths, so that even if not using "", the script will work smoothly and get all matching paths to the pattern. Small change that will remove some user confusion and frustration :)

asnyv:

---

### Contributor checklist

- [X] :tada: This PR closes #1096 .
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/smry2arrow_batch.py
+++ b/webviz_subsurface/smry2arrow_batch.py
@@ -27,7 +27,7 @@ def _get_parser() -> argparse.ArgumentParser:
         "\n"
         "Required input is a path/paths which may be wildcarded, like:\n"
         "  smry2arrow_batch my_folder/realization-*/iter-0\n"
-        "You may also define a spesific file pattern for the UNSMRY file to read."
+        "You may also define a specific file pattern for the UNSMRY file to read."
     )
 
     parser.add_argument(
--- a/webviz_subsurface/smry2arrow_batch.py
+++ b/webviz_subsurface/smry2arrow_batch.py
@@ -8,6 +8,7 @@ import logging
 import os
 import warnings
 from pathlib import Path
+from typing import List
 
 import ecl2df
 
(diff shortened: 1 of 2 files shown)


webviz_subsurface/smry2arrow_batch.py
```diff
@@ -24,20 +25,21 @@ def _get_parser() -> argparse.ArgumentParser:
     parser.description = (
         "Batch conversion of UNSMRY files to Apache Arrow IPC file format.\n"
         "\n"
-        "Note that if the enspath argument includes wildcards you will have to enclose it\n"
-        "in quotes to stop the shell from expanding it, e.g.:\n"
-        '  smry2arrow_batch "my_folder/realization-*/iter-0"'
+        "Required input is a path/paths which may be wildcarded, like:\n"
+        "  smry2arrow_batch my_folder/realization-*/iter-0\n"
+        "You may also define a spesific file pattern for the UNSMRY file to read."
```
anders-kiaer, line 30:
```suggestion
        "You may also define a specific file pattern for the UNSMRY file to read."
```

anders-kiaer (changes requested):

anders-kiaer:
A minor comment, otherwise LGTM! 👍

anders-kiaer (approved):

PR #1098, opened 2022-08-04
Support `dash>=2.6`

anders-kiaer:
It looks like the private class `NoUpdate` was moved in https://github.com/plotly/dash/commit/43e42eaecbd912c5391db5acd81566a35786a26a.

The class is only used in type checking on our side, so we have two options:
* Coninue use the dash private class in the type checking.
* Loosen up the type checking on our side.

After discussion, we go for the first solution, and hope dash itself gets type hinted in the near future (such that also `NoUpdate` becomes a public entity).

### DIFF
--- a/.github/workflows/subsurface.yml
+++ b/.github/workflows/subsurface.yml
@@ -49,7 +49,6 @@ jobs:
         if [[ $(pip freeze) ]]; then
           pip freeze | grep -vw "pip" | xargs pip uninstall -y
         fi
-        pip install "dash<2.6.0" # removed "_NoUpdate" in dash==2.6.0
         pip install "bleach<5"  # https://github.com/equinor/webviz-config/issues/586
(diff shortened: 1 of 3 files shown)


HansKallekleiv (approved):

setup.py
```diff
@@ -84,7 +84,7 @@
     },
     install_requires=[
         "flask-caching<1.11",  # https://github.com/equinor/webviz-subsurface/issues/1043
-        "dash>=2.0.0",
+        "dash>=2.6",
```
asnyv, line 87: Is it necessary to change this? the dash._callback NoUpdate existed in the 2.0.0 tag as well as far as I can see. 
https://github.com/plotly/dash/blob/2975e001cf017919929b0ebad1f1d1e14fa32f93/dash/_callback.py#L25

anders-kiaer: Good catch. Indeed it does (for some strange reason) - I'll keep the current lower limit then. Will fix.

PR #1107, opened 2022-08-23
Bugfix in `ParameterResponseCorrelation`

lindjoha:
Fixed a bug in `ParameterResponseCorrelation` that caused the plugin to fail if `response_filters` was not given as input parameter. Now the filters layout will always be added with at least the parameter filter.

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


asnyv:
Add to changelog please 😃

asnyv (commented):
With the `Filters` selector always there, you can simplify to a single list containing the `Controls` and `Filters` selectors without a `+` and so on

asnyv (approved):
👍

PR #1121, opened 2022-10-03
Fix `callback_typecheck` error for `wcc.Dialog` prop in `SimulationTimeSeries`

jorgenherje:
`open` is not required prop for wcc.Dialog, thus initial value was undefined, and the data type was undefined - i.e `None`. Resolved by defining initial state False, thus `open.propType = boolean` in `React`-component and python type is `bool`.


anders-kiaer (approved):
LGTM! 👍

PR #1122, opened 2022-10-03
Make it possible to `import webviz_subsurface` without `opm` and `ecl2df`. Include test install and import on `windows`

### LINKED ISSUE #520: not able to run webviz on mac
Trying to create some simple plugins however faced a crash on #webviz due to issue with #opm installation which is not possible to install on #mac , I am not planning to use any #opm related feature ... so in theory #webviz should not crash due to that ? 
image

asnyv:
- [X] :tada: This PR closes #520 (at least as far as we get for now, a lot of functionality is lost due to opm, ecl2df and fmu-ensemble not being compatible. It is thus possible to install `webviz-subsurface` and import `webviz_subsurface`, but the likelyhood of crashes due to using functionality that is dependent of these packages is **high**
   - [X] Wrap imports of `opm`, `ecl2df` and `fmu.ensemble` in `try/except`
- [X] :robot:  Added a test to install and import on `windows-latest` in `Github Actions` (also I let the `yaml` extension from `RedHat` in `vscode` autoformat the workflow file

### DIFF
--- a/.github/workflows/subsurface.yml
+++ b/.github/workflows/subsurface.yml
@@ -10,7 +10,7 @@ on:
       - published
   schedule:
     # Run CI daily and check that tests are working with latest dependencies
-    - cron:  '0 0 * * *'
+    - cron: "0 0 * * *"
 
 jobs:
   webviz-subsurface:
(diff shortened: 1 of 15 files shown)


asnyv:
shall we merge this @anders-kiaer ?

setup.py
```diff
@@ -98,7 +98,7 @@
         "pillow>=6.1",
         "pyarrow>=5.0.0",
         "pyscal>=0.7.5",
-        "scipy>=1.2",
+        "scipy>=1.2,<1.9.3",
```
anders-kiaer, line 101: Remove this now since it is an upstream issue, @lindjoha moved it into CI instead.

anders-kiaer (approved):
Yes, LGTM! Nice work. 👏 

One small comment below, should we also add a changelog entry? There have been a couple of requests on installing `webviz-subsurface` on Windows - this will enable that (with reduced functionality).

PR #1125, opened 2022-10-04
Option to remove header from parameter filter

lindjoha:
The header in the Parameter filter is not needed when it is used in a WLF settings group, because it has it's own name. This makes the header optional. But the default behavior is as before.

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


anders-kiaer (approved):
👍

PR #1130, opened 2022-10-05
New plugin: `VfpAnalysis`

lindjoha:
New plugin: `VfpAnalysis`. 

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


asnyv:
Is it ready for review? If so: could you fix the required testdata, add it to the test workflow (could be in a branch for now) and move this PR to "ready for review"? 😊

lindjoha:
> Is it ready for review? If so: could you fix the required testdata, add it to the test workflow (could be in a branch for now) and move this PR to "ready for review"? 😊

I have added a test file: tests/data/vfp.arrow, and a simple test.

webviz_subsurface/plugins/_vfp_analysis/_plugin.py
```diff
@@ -0,0 +1,52 @@
...
+class VfpAnalysis(WebvizPluginABC):
+    """Vizualizes VFP curves.
+
+    ---
+
+    * **`vfp_file_pattern`:** File pattern for where to search for vfp arrow files. The path
+    should be relative to the runpath if ensemble and realization is given as input, if not
+    the path needs to be absolute.
+    * **`ensemble`:** Which ensemble in `shared_settings` to use..
+    * **`realization`:** Which realization to pick from the ensemble.
```
anders-kiaer, line 19: In case the given realization has failed, this plugin will also then fail? Maybe ok given the speed increase you get with looking at a particular realization?

webviz_subsurface/plugins/_vfp_analysis/_views/_vfp_view/_utils/_vfp_figure_builder.py
```diff
@@ -0,0 +1,97 @@
+from typing import Any, Dict, List
+
+import plotly.graph_objects as go
+
+from ......_utils.colors import find_intermediate_color, rgba_to_str
```
anders-kiaer, line 5: 😄

anders-kiaer (approved):
Nice work! A couple of small comments only. 💬🚀

webviz_subsurface/plugins/_vfp_analysis/_plugin.py
lindjoha: Is it better to give a warning and open the app without any VFP curves? It currently will give a `FileNotFoundError` if the ensemble exists, but not the realization.

anders-kiaer: I think current approach with error instead of warning is preferred.

One alternative is perhaps to omit realization argument and instead use the first realization that succeeds (since we only take one realization we already assume it is constant across the ensemble)? But that has the downside of being less explicit... 🤔

I'm ok with both - you can choose @lindjoha 🙂

lindjoha: I think I'll just go for the current solution.

PR #1135, opened 2022-10-13
Metadata functionality in `EnsembleTableProvider`

lindjoha:
Implements the possibility to extract column metadata from the `EnsembleTableProvider`. It is needed in the `WellCompletion` plugin to get the KH unit.

Only unit is so far implemented, but it is easily extended to other column metadata

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/_providers/ensemble_table_provider/_field_metadata.py
+++ b/webviz_subsurface/_providers/ensemble_table_provider/_field_metadata.py
@@ -1,0 +1,24 @@
+from typing import Optional
+
+import pyarrow as pa
+
+from .ensemble_table_provider import TableVectorMetadata
+
+
+def create_vector_metadata_from_field_meta(
+    field: pa.Field,
+) -> Optional[TableVectorMetadata]:
+    """Create VectorMetadata from keywords stored in the field's metadata"""
+
+    meta_dict = field.metadata
+    if not meta_dict:
+        return None
+
+    try:
+        unit_bytestr = meta_dict[b"unit"]
+    except KeyError:
+        return None
+
+    return TableVectorMetadata(
+        unit=unit_bytestr.decode("ascii"),
+    )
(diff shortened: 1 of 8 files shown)


webviz_subsurface/_providers/ensemble_table_provider/_field_metadata.py
```diff
@@ -0,0 +1,24 @@
+from typing import Optional
+
+import pyarrow as pa
+
+from .ensemble_table_provider import TableVectorMetadata
+
+
+def create_vector_metadata_from_field_meta(
```
sigurdp, line 8: Currently, the `unit` property in `TableVectorMetadata` is defined as being optional, which implies that it is legal to return an empty `TableVectorMetadata` object. This function will however return `None` if metadata *is* present, but the `unit` key is missing.
If the current behavior of this function is what we want, the `unit` field in `TableVectorMetadata` should not be optional.
Alternatively, if the `unit` field should be kept as optional in `TableVectorMetadata`, this function should return an empty `TableVectorMetadata` object instead of returning `None` on line 20.

webviz_subsurface/_providers/ensemble_table_provider/ensemble_table_provider.py
```diff
@@ -1,9 +1,15 @@
 import abc
+from dataclasses import dataclass
 from typing import List, Optional, Sequence
 
 import pandas as pd
 
 
+@dataclass(frozen=True)
+class TableVectorMetadata:
```
sigurdp, line 9: Related to comment associated with `EnsembleTableProvider.vector_metadata()`, maybe reconsider the name of this class?

sigurdp (commented):
Added some comments related to method and class naming as well as behavior in the presence of partial meta data.

webviz_subsurface/_providers/ensemble_table_provider/_field_metadata.py
lindjoha: I see. My thinking behind making it `Optional` was to keep it flexible, in case a future use of `EnsembleTableProvider` has column metadata that is not `unit`. But then it is better to return an empty `ColumnMetadata` as you say.

Issue #774, opened 2021-09-21
GroupTree plugin: GRUPTREE/BRANPROP config setting


lindjoha:
Implement the option to choose whether to display the GRUPTREE or BRANPROP tree if both exists. One problem is that if a BRANPROP tree exists then the GRUPTREE nodes will not have group node pressures (GPR summary vectors).

lindjoha:
and we could also add the possibility of specifying the terminal node (could be FIELD by default). FIELD might not always be exported (for BRANPROP) and it would also be possible to show only a part of the tree.

PR #1138, opened 2022-10-18
More flexibility in `GroupTree`

### LINKED ISSUE #1136: `GroupTree`: options to specify `terminal_node` and `excl_well_startswith`
These features would be helpful to make the `GroupTree` more flexible and adaptable:

- Option to specify the terminal node (instead of always using FIELD)
- Option to filter out wells that starts with a given string (typically used to remove RFT wells)

lindjoha:
Improvements to the `GroupTree` plugin, making it more flexible and less strict on input.

---

### Contributor checklist

- [x] :tada: This PR closes #1136 and #774
   - [x] New input parameter `terminal_node`
   - [x] New input parameter `tree_type` that is `GRUPTREE` by default, but can also be set to `BRANPROP`. This is a breaking change since the plugin would previously use `BRANPROP` if that was available.
   - [x] New input parameters `excl_well_startswith` and `excl_well_endswith`
   - [x] Fixed bug related to `BRANPROP` entries for nodes with injection.
   - [x] Eased the requirements on which summary vectors must be present: FIELD production rates are not required, FIELD injection rates are only required if any wells have injection. Terminal node rates are not required and node data is not required (Pressure, BHP and WMCTL). This is partly because GRUPTREE trees don't have pressure if extended network is activated. Edge data (rates) are still required to make the tree consistant.
   - [x] tests for the functionality in `GruptreeModel`.
- [x] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.


anders-kiaer (approved):
Nice! 👏

PR #1141, opened 2022-10-29
`WellCompletion`: Get KH from arrow metadata instead of from parsing eclipse

### LINKED ISSUE #1134: `WellCompletion`: get KH unit from arrow file metadata
KH unit in the `WellCompletion` plugin should be taken from the` wellcompletiondata.arrow` metadata instead of from parsing eclipse files.

lindjoha:
---

### Contributor checklist

- [ ] :tada: This PR closes #1134 
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_subsurface/plugins/_well_completion/_utils/_well_completion_data_model.py
+++ b/webviz_subsurface/plugins/_well_completion/_utils/_well_completion_data_model.py
@@ -1,15 +1,11 @@
-import io
 import itertools
-import json
 import logging
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple
 
 import pandas as pd
-from webviz_config.webviz_store import webvizstore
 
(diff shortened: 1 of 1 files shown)


anders-kiaer (approved):
LGTM 👍 Add changelog entry? Does this require a recent version of generated arrow data from `ecl2df` (such that we should add this information in the plugin documentation), or is the `ecl2df` feature old enough that it would not be a problem in practice?

lindjoha:
> LGTM 👍 Add changelog entry? Does this require a recent version of generated arrow data from `ecl2df` (such that we should add this information in the plugin documentation), or is the `ecl2df` feature old enough that it would not be a problem in practice?

I'm thinking this is nothing the users will notice, and the feature in `ecl2df` is many months old now. I think there was only one komodo release where it was not there 2022.07 maybe. So I think it will not be a problem.

PR #1147, opened 2022-11-04
Modify SimulationTimeSeries to get data from Sumo

alifbe:
This PR is used to show modifications in SimulationTimeSeries to read data from Sumo instead of scratch disk


alifbe:
@anders-kiaer I have pushed the code now. Unfortunately, it will not pass the checks because the plugin require different input parameters than the  original requires.

Issue #1154, opened 2022-11-08
Fix `no_implicit_optional=True` for `mypy >= 0.990`


jorgenherje:
`mypy` with version `>= 0.990` has changed its default to `no_implicit_optional=True`. This implies error for all missing `Optional` or `Union[..., None]` where default value is set to `None`.

Temporary solution is to set `implicit_optional=True` in `mypy.ini` as plugin `__init__`-functions cannot handle usage of `Optional` when JSON Schema is generated using typehints runtime.

PR #1155, opened 2022-11-08
Set `implicit_optional=True` for `mypy`

jorgenherje:
`mypy` with version `>= 0.990` has changed its default to `no_implicit_optional=True`. This implies error for all missing `Optional` or `Union[..., None]` where default value is set to `None`.

Temporary solution is to set `implicit_optional=True` in `mypy.ini` as plugin `__init__`-functions cannot handle usage of `Optional` when JSON Schema is generated using typehints runtime.

Closes: #1154


anders-kiaer (approved):
👍 LGTM

Issue #1106, opened 2022-08-18
Remove direct import from `SimulationTimeSeries` into other plugins


jorgenherje:
Plugins as `_bhp_qc.py`, `_prod_misfit` and `_property_statistics` imports directly from `SimulationTimeSeries` plugin. 

- `ProviderSet` should be moved to higher folder structure for reusability among plugins. 
- Consider `datetime_utils` `from_str()` method to be a common utility? Utilized in `_property_statistics`

---

Discuss if  `ProviderSet` class and corresponding utilities are general enough for reusability?

`ProviderSet`: https://github.com/equinor/webviz-subsurface/blob/master/webviz_subsurface/plugins/_simulation_time_series/types/provider_set.py

`datetime_utils`: https://github.com/equinor/webviz-subsurface/blob/master/webviz_subsurface/plugins/_simulation_time_series/utils/datetime_utils.py

PR #1161, opened 2022-11-11
Remove direct import from `SimulationTimeSeries` utilities

jorgenherje:
Various plugins performs direct import of utility objects/functions from `SimulationTimeSeries`.

Move the respective utils to a common top-level utility folder. Utilities which are too general, and does not belong in the top-level utility folder, are moved into the specific plugins to prevent misusage.

---

#### Done:

- Moved the `ProviderSet` to `webviz_subsurface/_utils` for common usage.
    - Renamed ` ProviderSet` to `EnsembleSummaryProviderSet` as the utility is now moved out of scope of `SimulationTimeSeries`. There are multiple `providers` in `webviz-subsurface` and a common utility must clarify its scope - i.e. set for `EnsembleSummaryProvider`.
    - Note: 
        - `names()`-method is renamed to `provider_names()` for clarity.
        - The provider set functionality is assuming same implementation of `EnsembleSummaryProvider` interface across all provides in set.
- Moved the `ProviderSet` creator methods to `webviz_subsurface/_utils` 
    - File renamed to `ensemble_summary_provider_set_factory.py` according to renaming to `EnsembleSummaryProviderSet`
- Removed `webviz_subsurface/_utils/datetime_utils.py`
   - Was a direct copy of utility in `SimulationTimeSeries`-plugin folder. The implementations assumes data with daily frequency, thus the date conversion `to_str` and `from_str` methods can result in loss of data if date frequency is higher than daily. This should not be common utility methods. File deleted and utility methods added directly in respective plugins. 
- Updated of imports according to new file placement
- Updated/moved unit tests to new folders according to new naming and placement

Closes: #1106


anders-kiaer (approved):
Very good :clap: LGTM! :100:

PR #1170, opened 2022-11-18
Include `vtk` dependencies in Docker example app

anders-kiaer:


asnyv (commented):
seems like you probably have to prepend the includes with a line shift?

anders-kiaer:
Hmm, yes, strange, it did work without explicit line shift on my computer™. Maybe small variations in how shells interpret `>>`.

asnyv (approved):

Issue #1164, opened 2022-11-17
Fix incorrect use of ensemble providers


asnyv:
**Describe the bug**
Some plugins use the new providers in a memory inefficient way, loading data from the providers and storing the output in the plugins attributes in `init` as e.g. `DataFrame`, resulting in the same memory issue we originally intended to mitigate. 

**Expected behavior**
Data should be fetched from the provider on demand, and not kept in the state of the plugin instance.

**Additional context**
A bit hard to find all occurances of misconfigured providers, so think it is better if each tries to remember where it has been implemented, and then we can maybe do a screening of memory usage to find places we have missed.

Some examples where summary data is stored directly as part of the plugin instance:
- https://github.com/equinor/webviz-subsurface/blob/a077e475c2fdbd8124821b780e696135f119b69d/webviz_subsurface/plugins/_well_analysis/_utils/_ensemble_well_analysis_data.py#L42
- https://github.com/equinor/webviz-subsurface/blob/e94ca641a691d262ee42c244470bf337dc2f379d/webviz_subsurface/plugins/_bhp_qc.py#L70
- https://github.com/equinor/webviz-subsurface/blob/b9a29e31c400291f90db0e93ca91dfbb3c83ffb4/webviz_subsurface/plugins/_parameter_response_correlation.py#L195

Example where tabular data is stored direclty as part of the plugin instance (technically later used :
- https://github.com/equinor/webviz-subsurface/blob/b9a29e31c400291f90db0e93ca91dfbb3c83ffb4/webviz_subsurface/plugins/_parameter_response_correlation.py#L171-L207

PR #1182, opened 2022-12-05
ParameterResponseCorrelation empty column_keys

### LINKED ISSUE #1181: Allow empty column_keys in ParameterResponseCorrelation plugin when using arrow file
We are trying to switch from CSV file to arrow file for ParameterResponeCorrelation plugin. When using CSV file as input, user does not need to specify column_keys and the plugin will load all columns in the CSV. When using arrow file, column_keys becomes compulsory

image

Allow users to not specify column_keys in advance and let plugin load all existing columns.

alifbe:
When we switch from csv file `response_file` to arrow file `rel_file_pattern`, we get an error which basically requires us to specify the column_keys instead of set it to default `None` 

Specifying all the column_keys in advance could be tedious and in-practical in some cases. So, it is desired to keep the same behavior (load all columns) as when we are using csv file as input.

---

### Contributor checklist

- [x] :tada: This PR closes #1181.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [UNRELEASED] - YYYY-MM-DD
 
 ### Changed
+- [#1182](https://github.com/equinor/webviz-subsurface/pull/1182) - `ParameterResponseCorrelation` can now allow default None column_keys when using arrow file as input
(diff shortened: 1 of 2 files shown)


alifbe:
@anders-kiaer can you please check this PR?

asnyv (commented):
Thanks 😊
We have noted a bigger issue with the implementation in general (see #1164), but good with a short term fix for your issue. Could you add an entry to the changelog as well?

alifbe:
> Thanks blush We have noted a bigger issue with the implementation in general (see #1164), but good with a short term fix for your issue. Could you add an entry to the changelog as well?

Thanks for checking @asnyv 

I have now updated the changelog.md

alifbe:
Thanks @asnyv 

I don't have write access to the repo. So it will be great if you can merge the branch 👍

PR #1183, opened 2022-12-06
Support lazy loading of individual plugins

anders-kiaer:
Utilize the new possibility in plugin projects after https://github.com/equinor/webviz-config/pull/656.


anders-kiaer:
Blocked by https://github.com/equinor/webviz-config/issues/660.

Issue #1163, opened 2022-11-17
Place the `VectorCalculator` in `SimulationTimeSeries` in a `WebvizDialog`


jorgenherje:
Replace the `wcc.Dialog` with `wcc.WebvizDialog` for the `VectorCalculator` in the `SimulationTimeSeries`-plugin

jorgenherje:
Outdated

PR #1186, opened 2022-12-19
Place `VectorCalculator` in SimulationTimeSeries-plugin in new `WebvizDialog`-component

jorgenherje:
Place `VectorCalculator` in `SimulationTimeSeries`-plugin into the new `WebvizDialog`-component from `webviz-core-components`.

Closes : #1163


jorgenherje:
Outdated

PR #1193, opened 2023-02-01
Use view annotations in mapviewerfmu

HansKallekleiv:
Fix `DeckGLMap` renamed to `DashSubsurfaceViewer`.

Use the `children` prop to display view annotations for label and color legends.


anders-kiaer (approved):
🚀

Issue #1007, opened 2022-04-22
Handling of csv polygon files


rnyb:
Currently there's a couple of webviz plugins using polygon files (csv format); SeismicMisfit and RftPlotter (others?).
These plugins expects that the header = `X_UTME, X_UTMN, Z_TVDSS, POLY_ID`
However, we should consider supporting a new header standard = `X, Y, Z, ID`

For a period of time we could implement support for both versions (with a deprecation warning on the old?).

We could also consider adding a new class for reading polygon files for re-use across plugins

rnyb:
Plugins to update

- [ ] SeismicMisfit
- [ ] RftPlotter
- [ ] VolumetricAnalysis

PR #1195, opened 2023-02-21
Update ``RftPlotter`` so it works with new header for polygon files.

rnyb:
By default, datatio writes X, Y, Z, ID as header names. This used to be X_UTME, Y_UTMN, etc. This implementation makes the plugin work with both headers.

### Contributor checklist

- [x] :tada: This PR is related to #1007 


tnatt (approved):
👍

PR #1201, opened 2023-03-24
`ParameterAnalysis` to `WLF`

### LINKED ISSUE #567: Add Python type hints for ParameterAnalysis plugin
We should add type hints and enable `mypy` type checking for the `ParameterAnalysis` plugin.

This plugin will be involved in the redesign and consolidation of `SimulationTimeSeriesModel` and we should make sure we have mypy support in place before embarking on this.

lindjoha:
This PR converts the `ParameterAnalysis `plugin to WLF. 

The functionality is mostly the same, but some bugs have been fixed f.ex related to summary vectors present in one ensemble and not in the other. Mypy is now used in the CI workflow for this plugin and the code quality is therefore improved many places.

Another change is that the plugin does not try to categorize ensembles as MonteCarlo or Sensitivity ensembles. All ensembles are loaded into the parameter response view, and it's up to the user to make sense out of it. The plan is to add a sensitivity filter in the near future.

---

### Contributor checklist

- [x] :tada: This PR closes #567 .
   - [ ] Task 1
   - [ ] Task 2


tnatt (approved):
LGTM 👍 
liked the look of the plugin with the WLF! 
minor things only:
- we should increase the view height of the parameter distribution plot to be equal to the menu 
- and maybe set the correlation view as the default start, it is a bit hidden now..

image

PR #1212, opened 2023-04-24
Pvt issue 1202

### LINKED ISSUE #1202: Bugs in `PvtPlot`
A few apparent bugs in the `PvtPlot` plugin.

The GOR plot is in fact plotting density, not GOR., see: https://github.com/equinor/webviz-subsurface/blob/3d76d6b969c8059894c6d7dfcb8129e0460e7da5/webviz_subsurface/plugins/_pvt_plot/_views/_pvt/_utils/_plot_utils.py#L267-L274

The GOR plot lacks `name` defined for the trace, thus giving it a simple "trace 0" and "trace 1" in the legend: 
image

Also seems like there is some mix-up with saturated properties caused by a sorting as it is followed by iloc indexing. Not sure if this sorting is really needed? https://github.com/equinor/webviz-subsurface/blob/3d76d6b969c8059894c6d7dfcb8129e0460e7da5/webviz_subsurface/plugins/_pvt_plot/_views/_pvt/_utils/_plot_utils.py#L92-L95 

Might be useful with some help from @rubenthoms if you still remember any of the code base for the plugin 🤔 especially whether that sorting is necessary for something. I have already done some debugging and tests of changes locally. Will open a draft PR once I've cleaned it up a bit

asnyv:

---

### Contributor checklist

- [ ] :tada: This PR closes #1202.
   - [ ] Task 1
   - [ ] Task 2


webviz_subsurface/plugins/_pvt_plot/_views/_pvt/_utils/_plot_utils.py
```diff
@@ -89,10 +89,10 @@ def create_traces(
...
         dim_column_name = "PRESSURE"
 
-    data_frame = data_frame.sort_values(
-        ["PRESSURE", "VOLUMEFACTOR", "VISCOSITY"],
-        ascending=[True, True, True],
-    )
+    # data_frame = data_frame.sort_values(
+    #     ["PRESSURE", "VOLUMEFACTOR", "VISCOSITY"],
+    #     ascending=[True, True, True],
+    # )
```
rubenthoms, line 95: If I remember correctly this was only to ensure that the list is sorted and that the points plotted are connected with lines in the correct order (under the assumption that Plotly is plotting in the order given by the input array). This might not have an effect since the data frame might already be sorted beforehand. However, I wonder if the plots would still be correct if the df wasn't. If we are 100% sure that the df is always sorted correctly, this can be removed.

rubenthoms (approved):
LGTM. There is still uncertainty if the data frame is coming in presorted.

anders-kiaer (approved):
As discussed offline, we merge this as is and include it in the upcoming komodo test phase. We think the correct way is to use the incoming sorting of the Eclipse PVT data (i.e. the way it comes out of `ecl2df`).

PR #1213, opened 2023-04-27
Support rendering of surfaces as images in MapViewerFMU

HansKallekleiv:
The new `MapLayer` for surface visualization is too slow for FMU work.
Added a new config flag `render_surfaces_as_images:bool` to allow the old image based behaviour.
Also setting the flag as default to `True`.


tnatt (approved):
LGTM 👍 the performance is much better now, and nice to finally have the option of truncating the colorscale! 

it gets a bit tight with the data min and max displayed.. could it be available only on hover maybe? and increase the size of the input fields for min and max to the same width as the "reset range button"? 
image

also the new colorscales added are nice but many options are very similar.. should we limit them a bit?
image
image

PR #1214, opened 2023-05-02
Bugfiz in `ParameterAnalysis`

lindjoha:
When the selected vector was also included in the list of vectors for parameter correlation then it was duplicated in the list passed to the summary provider. This gave an error. F.ex if the selected vector is `FOPT` and the correlation vector is `F*`. The solution is to remove duplicates from the list.

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


anders-kiaer (approved):
👍

Issue #1224, opened 2023-06-16
Sensitivity filter updates time series but not tornado plot


anders-kiaer:
![image](https://github.com/equinor/webviz-subsurface/assets/31612826/f72b49c0-bc60-4046-9c0a-b9fba4a1ea72)

tnatt:
Closed in https://github.com/equinor/webviz-subsurface/pull/1227

PR #1227, opened 2023-07-17
Add filter linking in `SimulationTimeseriesOneByOne`

tnatt:
PR that adds the option to use the sensitivity filter on all elements or only the timeseries graph in the `SimulationTimeseriesOneByOne` plugin.
- deafault behaviour is set to use the filter across all visualisations

Solves #1224

image


lindjoha (approved):
Very good :)

PR #1230, opened 2023-08-01
Fix CI and enable mypy again

tnatt:
PR that fixes the failing CI tests, and enables mypy in the CI again.

Tests were failing due to pyscal removing the dependencies on scipy<1.93 https://github.com/equinor/pyscal/pull/401. This led Scipy to be updated, using an improved interpolation algorithm which led to the assertion error. 
It was also needed to unpin the xtgeo version to include a bugfix https://github.com/equinor/xtgeo/pull/822


asnyv (approved):
👍

PR #1234, opened 2023-08-17
Bugfixes in the `EnsembleTableProvider` and `SimulationTimeseriesOneByOne`

### LINKED ISSUE #1228: `SimulationTimeSeriesOneByOne`: unexpected behavior when changing ensembles with different sensitivity names
When the `SimulationTimeSeriesOneByOne` plugin is used on two ensembles with different sensitivity names, the behavior when changing ensembles is not good. All the sensitivities is de-selected from the select-box and the filter doesn't seem to be applied at all. 

When selecting new sensitivities the behavior is as expected.

The wanted behavior is probably that "new" sensitivity names is selected by default when changing ensembles.

image

tnatt:
PR with some bugfixes:

- Bugfix in the `EnsembleTableProvider` to allow spaces in parameter values, when loading parameters.txt. (e.g in a sensitivity name). 
  - This change is done on a unreleased feature, hence no changelog entry.

- Bugfixes in the `SimulationTimeseriesOneByOne` 
  - Change selected sensitivity filter values when ensemble is changed. Closes #1228
    - Here the new feature in Dash to allow for duplicate outputs in callbacks has been utilized to avoid combining two callbacks into one.
  - Prevent update if no sensitivities have been selected in the sensitivity filter
  - Avoid Error if no sensitivity ensembles are input to the plugin
    - handled by allowing existing functionality in the `ParametersModel `to insert values for `SENSNAME/SENSCASE`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - [#1232](https://github.com/equinor/webviz-subsurface/pull/1232) - Removed slow function in `RftPlotter` to improve startup time of plugin.
+- [#1234](https://github.com/equinor/webviz-subsurface/pull/1234) - Fixed a bug that occured in `SimulationTimeSeriesOneByOne` when changing between ensembles with different sensitivity names.
+
 
 ### Changed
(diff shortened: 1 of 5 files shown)


lindjoha (approved):
LGTM!

PR #1244, opened 2023-11-15
Added facies fractions to `VolumetricAnalysis` and smaller bugfixes

tnatt:
PR that adds new functionality to compute facies fractions per individual facies in the `VolumetricAnalysis` plugin.
To compute facies fractions, FACIES needs to be included as a groupby selector. To inform users text is outputted to the screen.
The fractions is calculated as the sum of bulk for a facies divided by the total bulk sum - this gives exact results as weighting on bulk volumes inside RMS.

Other smaller improvements in this PR:
- added text label to bar plots, and possibility to control the number formatting from settings. Note the Default option will try to guess the format based on the selected responses.
- added more fluid representative coloring to plots if the "FLUID_ZONE" is used as Color by. 

image
image

Other smaller bug fixes in this PR:
- ensure that the boxmode is always valid for box plots
- ensure the plot type is never None when the page is "Custom plotting"
- changed the check for Water volumes computation to only check the required "BULK", "NET", and "PORV" columns.
- removed duplicate "Reference" column in tornado table, it is now given as default by the `TornadoTable`

### DIFF
--- a/webviz_subsurface/plugins/_volumetric_analysis/utils/table_and_figure_utils.py
+++ b/webviz_subsurface/plugins/_volumetric_analysis/utils/table_and_figure_utils.py
@@ -5,6 +5,15 @@ import plotly.graph_objects as go
 import webviz_core_components as wcc
 from dash import dash_table
 
+from webviz_subsurface._models import InplaceVolumesModel
+from webviz_subsurface._utils.colors import StandardColors
+
+FLUID_COLORS = {
+    "oil": StandardColors.OIL_GREEN,
+    "gas": StandardColors.GAS_RED,
+    "water": StandardColors.WATER_BLUE,
+}
+
 
 def create_table_columns(
     columns: list,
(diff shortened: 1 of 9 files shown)


webviz_subsurface/plugins/_volumetric_analysis/utils/table_and_figure_utils.py
```diff
@@ -5,6 +5,14 @@
 import webviz_core_components as wcc
 from dash import dash_table
 
+from webviz_subsurface._models import InplaceVolumesModel
+
+FLUID_COLORS = {
+    "oil": "#007079",
+    "gas": "#FF1243",
+    "water": "#ADD8E6",
+}
```
lindjoha, line 14: Use the type `StandardColors` in `_utils/colors.py`  ?

lindjoha (approved):
LGTM

PR #1248, opened 2023-11-29
Adjust naming according to new property names after splitting GroupTree-component view/settings

jorgenherje:
Adjustments to match: https://github.com/equinor/webviz-subsurface-components/pull/1794

----

Await new release of `wsc` before finalizing this PR.


anders-kiaer (approved):
👍 Good timing for taking this in now wrt. komodo release cycle.

PR #1266, opened 2024-03-11
Add volume change analysis plot to VolumetricAnalysis

tnatt:
PR which resolves #1264, and adds waterfall plots for analyzing a volumetric change to the existing `comparison` tabs in `VolumetricAnalysis` .

The waterfall plot functionality is only supported for static volumes, and will only work for STOIIP or GIIP response.
If a `NET` column is present in the dataset `NTG` and `PORO_NET` will be used in the waterfall plot otherwise the `PORO `column is used.

The waterfall functionality will work both for `ensemble`, `source` and `sensitivity` volumes comparisons.

image

PR #1268, opened 2024-03-18
Add changelog entry for waterfall plot

tnatt:
Add changelog entry for #1266.

Also reduced the line lengths of the radio button text, to avoid text wrapping. They were a bit long after the move to selections.


rnyb (approved):
👍

Issue #1281, opened 2024-04-30
Fix pylint reported conventions and refactor proposals


andreas-el:
From this step:
https://github.com/equinor/webviz-subsurface/actions/runs/8894159783/job/24421831320

Note that this PR lowers the pylint standards ever so slightly to allow builds to pass again.
This should be adjusted back to `10.0` once these issues have been resolved.

ref.
https://github.com/equinor/webviz-subsurface/pull/1277

Issue #1255, opened 2024-01-24
Support Python3.11


larsevj:
With the ongoing movement towards RHEL8 and Python3.11 for komodo, please consider adding Python3.11 support.

asnyv:
depends on https://github.com/equinor/webviz-config/pull/675

eivindjahren:
We are starting to build with python 3.11, but run into this problem: 
```
webviz-subsurface 0.2.25.dev6+gdef9a93 has requirement opm<=2022.10,>=2020.10.1; sys_platform == "linux", but you have opm 2023.10.
webviz-subsurface 0.2.25.dev6+gdef9a93 has requirement pandas<2.0,>=1.1.5, but you have pandas 2.2.1.
```

Issue #1258, opened 2024-01-26
`opm` `2023.10` do not work with our Docker base image


anders-kiaer:
When importing `opm` in the Docker image it complains about
```
libgomp.so.1: cannot open shared object file: No such file or directory
```

PR #1282, opened 2024-04-30
Support Python 3.11 & unpin OPM

andreas-el:
Resolves https://github.com/equinor/webviz-subsurface/issues/1255
Resolves https://github.com/equinor/webviz-subsurface/issues/1281
Resolves https://github.com/equinor/komodo-releases/issues/5156

Unpins OPM, but sets pin for python `3.8` due to newer versions not supporting that:
```
"opm>=2020.10.1,<2023.10; python_version=='3.8' and sys_platform=='linux'",
"opm>=2023.10; python_version>'3.8' and sys_platform=='linux'",
```

Introduces a shim used in ERT, where versions prior to `3.11` retain enum functionality, while newer versions use `StrEnum` instead.

Resolves a lot of pylint conventions, refactors and information issues.


eivindjahren (approved):
Just reorder the commits and we are good to go.

anders-kiaer (approved):
LGTM. Thank you 🙇 Not directly related to this PR: I think there is some issue with the latest OPM binaries wrt. how they run in the Docker Ubuntu image (https://github.com/equinor/webviz-subsurface/issues/1258), but we will be able to fix that as a follow-up PR.

PR #1286, opened 2024-05-07
Provide pandas 2 compatibility

andreas-el:
Resolves https://github.com/equinor/webviz-subsurface/issues/1288

Build using pandas 2.0.2/2.2.2:
https://github.com/equinor/webviz-subsurface/actions/runs/9110545525

Build using pandas 1.5.3:
https://github.com/equinor/webviz-subsurface/actions/runs/9110407584

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,7 @@ setup(
         "geojson>=2.5.0",
         "jsonschema>=3.2.0",
         "opm>=2023.10; sys_platform=='linux'",
-        "pandas>=1.1.5,<2.0",
+        "pandas>=1.1.5",
         "pillow>=6.1",
         "pyarrow>=5.0.0",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -98,7 +98,7 @@
         "geojson>=2.5.0",
         "jsonschema>=3.2.0",
         "opm>=2023.10; sys_platform=='linux'",
-        "pandas>=1.1.5,<2.0",
```
eivindjahren, line 101: I just checked and webviz-subsurface is not compatible with `pandas-3.0.0dev0` so to keep the wheels spinning I suggest we have `pandas>=1.1.5,<3`.

eivindjahren: :+1:

HansKallekleiv (approved):

PR #1287, opened 2024-05-08
Fix bug when grouping on FACIES for non-standard static sources

tnatt:
PR to fix a bug where users would get an KeyError on `BULK`  if the csv-file contained `FACIES` but did not contain `BULK`. 
Typically this would happen for non-standard static sources where `BULK_OIL` and `BULG_GAS` columns are kept.


rnyb (approved):
lgtm

andreas-el:
I think it makes sense to pin pylint < `3.2.0` to avoid all this for the time being.
I have done so in my current PR, and already created an issue for resolving the pylint issues down the line.
Ref; https://github.com/equinor/webviz-subsurface/issues/1292

tnatt:
> I think it makes sense to pin pylint < `3.2.0` to avoid all this for the time being. I have done so in my current PR, and already created an issue for resolving the pylint issues down the line. Ref; #1292

Thanks for letting me know @andreas-el! Yes, I saw your issue briefly and figured I just wait it out and see if you were going to address it soon 😄 But then I'll just pin it instead for the time being 👍

PR #1293, opened 2024-05-21
Add column with total oil/gas volumes to VolumetricAnalysis

tnatt:
Add automatic calculation of `STOIIP_TOTAL` and `GIIP_TOTAL` columns to `VolumetricAnalysis`.

Drogon example:
image

Also included a bugfix on the logic for evaluating if the data is compatible for a water fall plot

### DIFF
--- a/webviz_subsurface/_models/inplace_volumes_model.py
+++ b/webviz_subsurface/_models/inplace_volumes_model.py
@@ -36,6 +36,8 @@ class InplaceVolumesModel:
         "GIIP",
         "ASSOCIATEDOIL",
         "ASSOCIATEDGAS",
+        "STOIIP_TOTAL",
+        "GIIP_TOTAL",
         "BULK",
         "NET",
         "PORV",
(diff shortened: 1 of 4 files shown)


webviz_subsurface/_models/inplace_volumes_model.py
```diff
@@ -93,6 +95,16 @@ def __init__(
                 self._dataframe, self.pmodel.sens_df, on=["ENSEMBLE", "REAL"]
             )
 
+        # create HC_TOTAL columns
+        if "STOIIP" in self._dataframe and "ASSOCIATEDOIL" in self._dataframe:
+            self._dataframe["STOIIP_TOTAL"] = self._dataframe["STOIIP"].fillna(
+                0
```
rnyb, line 101: is the `fillna(0)` split needed here?

webviz_subsurface/_models/inplace_volumes_model.py
```diff
@@ -93,6 +95,16 @@ def __init__(
...
+            ) + self._dataframe["ASSOCIATEDOIL"].fillna(0)
+        if "GIIP" in self._dataframe and "ASSOCIATEDGAS" in self._dataframe:
+            self._dataframe["GIIP_TOTAL"] = self._dataframe["GIIP"].fillna(
+                0
```
rnyb, line 105: ditto

rnyb (approved):
Nice addition. Some micro comments only.
Could there be a use case  to also have the total column included when there is only one of the stoiip columns is present?

webviz_subsurface/_models/inplace_volumes_model.py
tnatt: Yes it is due to e.g. the `STOIIP` being `nan` for `FLUID_ZONE != "oil"`.. so need to fill it before adding them together otherwise the resulting `STOIIP_TOTAL` is also `nan` 🙂

PR #1317, opened 2025-02-07
Fix ensemble colors in relperm

HansKallekleiv:
Change in Pandas: When grouping on a list of columns a tuple is returned. Previously just a string was returned as long as the list only had 1 item.
This caused the ensemble name to be a tuple, which again broke the legend / coloring in plotly
image


asnyv (approved):
👍

PR #1324, opened 2025-03-20
Make legend state persist when changing graph settings

vegardkv:
Fixes issue XXX

Follows same principle as one of the comments in _plugin.py from "TwoProductiveDays" branch, except that the legend state is stored via dcc.Store instead.

Toggling legend items on/off should now be "remembered" for when settings change


vegardkv:
PR created in wrong repo...

PR #1327, opened 2025-04-14
Add histogram to parameter distributions

### LINKED ISSUE #1323: Add histogram option
The current only option (besides box plot) is to show kernel/density functions. This type of plot "smoothens" the underlying data, making it a bit harder to understand e.g. min-max and exactly how the different realizations fall in different bins/ranges.

Add histogram option (and make this the default?)

Image

HansKallekleiv:
Closes #1323 
Plotly calculates histogram bins automatically.  The default behaviour is shared binning across all subplots.
This is avoided by setting `bingroup=None`. Bins are then calculated per trace.
This means that ensembles within each subplot also has separate binning.

image


anders-kiaer (approved):
LGTM 👍

PR #1328, opened 2025-04-14
Use single color for distribution sensitivities in Tornado plots

HansKallekleiv:
Use same color for p10/p90 "cases" when calculated from a distribution.

Should we also show realizations points as default?

image
image


anders-kiaer:
I vote for showing realization points as default, but not a strong opinion.

anders-kiaer (approved):

PR #1333, opened 2025-05-02
Handle parameters from parameters.txt with DESIGN_MATRIX prefix

HansKallekleiv:
Support new `DESIGN_MATRIX` prefix in ERT.
This is a simple fix, just removing the prefix on loading of the parameters file so that further logic works as before.
Added a check for duplicates after the renaming, dropping any parameters without the prefix if it would results in duplicates.


anders-kiaer (approved):
LGTM 👍

PR #1371, opened 2026-06-26
Support new `dash` versions

anders-kiaer:
Current `dash<=2` limit makes it impossible to avoid some dependency vulnerabilities.

Allowing `dash>=4` involves changing e.g. `DataTable` import, as well as changes to types in order to satisfy linters.

Tested on https://webviz-subsurface-example.azurewebsites.net/ (current version there is with dash 4).


anders-kiaer:
Superseded by #1373

